PR #229, opened 2023-01-11
Add new HEAD method to return entry metadata without reading entire record

### LINKED ISSUE #214: Feature:  HEAD methods to read only meta information about record
### Current problem

Sometimes we need only check if a record with a certain timestamp exits or figure out its size. We don't need to read the whole content for it and may take all the information from HTTP headers.

### Possible solution

We need an additional endpoint `HEAD /api/v1/:bucket/:entry/`which works exactly like the `GET`method but doesnt send the record content 

### Additional context

_No response_

### Code of Conduct

rtadepalli:
Closes #214 

- [X] Tests for the changes have been added (for bug fixes / features)
- [ ] CHANGELOG.md have been updated (for bug fixes / features / docs)

This PR introduces a new HTTP endpoint of type HEAD that returns entry metadata without reading the entire record. It is functionally similar to the GET endpoint, except we return everything else except the record content.

(Bug fix, feature, docs update, ...)

This API does not exist right now, this is a new feature.

### What is the new behavior?

This is a new feature.

No, there are no breaking changes because of this PR.


rtadepalli:
I will update documentation once the core changes look satisfactory.

api_tests/entry_api_test.py
```diff
@@ -327,3 +327,34 @@ def test__query_with_read_token(base_url, session, bucket, token_without_permiss
 
     resp = session.get(f'{base_url}/b/{bucket}/entry/q', headers=auth_headers(token_write_bucket))
     assert resp.status_code == 403
+
+
+def test__head_entry_ok(base_url, session, bucket):
+    ts = 1000
```
atimin, line 354: Some little doc string is missing. Something like "should return only headers"

api_tests/entry_api_test.py
```diff
@@ -327,3 +327,34 @@ def test__query_with_read_token(base_url, session, bucket, token_without_permiss
...
+    dummy_data = "dummy data"
+    resp = session.post(f'{base_url}/b/{bucket}/{entry_name}?ts={ts}', data=dummy_data)
+    assert resp.status_code == 200
+
+    resp = session.head(f'{base_url}/b/{bucket}/{entry_name}')
+    assert resp.status_code == 200
+    assert len(resp.content) == 0
+    assert resp.headers['Content-Length'] == str(len(dummy_data))
+
+
```
atimin, line 365: I would like to add a test for error. Ask, please, for a non-existing record and check "x-reduct-error"

src/reduct/api/entry_api.cc
```diff
@@ -175,12 +175,14 @@ Result<HttpRequestReceiver> EntryApi::Read(IStorage* storage, std::string_view b
                 .headers = std::move(headers),
                 .content_length = reader->size(),
                 .SendData =
-                    [reader]() {
-                      auto [chunk, err] = reader->Read();
-                      return Result<std::string>{std::move(chunk.data), err};
+                    [reader, send_record_data]() {
```
atimin: this is completely fine to pass "send_record_data into the lambda. Consider using the flag to create two different lambdas.

atimin (changes requested):
Thank you for your contribution. Almost perfect! Please see my comments and check the pipeline.   `cpplint` isn"t happy ;).  You can use `pre-commit` to check the format automatically when you make a commit.

atimin (commented):
Unfortunately, I found another problem. Check http_server.cc:219:

```cpp
 auto [ok, responded] =
            ctx.res->tryEnd(chuck.substr(ctx.res->getWriteOffset() - offset), response.content_length);
        if (ok) {
          complete = responded;
          break;
```

We use the `content_length` for sending data. When we send an empty string but say the size is greater than 0, the HTTP connection hangs.  We should work around the case for the HEAD request. 

WebSocket framework sets the "Content-Length" header automatically, not sure if it is possible to send an empty trying but have the non-zero header. I'll check myself

atimin:
Blocked by uNetworking/uWebSockets#1557

atimin:
@rtadepalli I've merged it to keep your contribution after Rust migration.

PR #231, opened 2023-01-16
Store the content-type received in the header into the record metadata

### LINKED ISSUE #219: Feature: Keep original type of record
### Current problem

Currently, the database ignores the `content-type` header of an HTTP request when it writes a record.  When a client reads a record it receives it as `application/octet-stream`.  In many cases, the content type is useful information, we should keep it. 

### Possible solution

We could add the `content_type` field  to the `Record`  message see (src/reduct/proto/storage/entry.proto) and use instead of `application/octet-stream` (see  `reduct::api::EntryApi::Read`)

### Additional context

_No response_

### Code of Conduct

rtadepalli:
Closes #219 

- [X] Tests for the changes have been added (for bug fixes / features)
- [X] Docs have been added / updated (for bug fixes / features)
- [X] CHANGELOG.md have been updated (for bug fixes / features / docs)

If a record is written with the `Content-Type` header, this should be preserved so that the record can be returned with the same header later.

All records are returned using the `Content-Type: application/octet-stream` header.

### What is the new behavior?

Stores the `Content-Type` header in the record metadata.

No, this does not introduce a breaking change.

### DIFF
--- a/src/reduct/proto/storage/entry.proto
+++ b/src/reduct/proto/storage/entry.proto
@@ -19,7 +19,7 @@ message Record {
   }
 
   message Label {
-    string name =1;
+    string name = 1;
     string value = 2;
   }
 
@@ -28,6 +28,7 @@ message Record {
   uint64 end = 3;                           // end position of a blob in a block
   State state = 4;                          // state of record
   repeated Label labels = 5;                // labels
+  optional string content_type = 6;         // Content type of the record
 }
 
 // Represents a block of records.
--- a/src/reduct/proto/storage/entry.proto
+++ b/src/reduct/proto/storage/entry.proto
@@ -28,7 +28,7 @@ message Record {
   uint64 end = 3;                           // end position of a blob in a block
   State state = 4;                          // state of record
   repeated Label labels = 5;                // labels
-  optional string content_type = 6;         // Content type of the record
+  string content_type = 6;                  // Content type of the record
 }
 
 // Represents a block of records.
--- a/src/reduct/storage/io/async_reader.cc
+++ b/src/reduct/storage/io/async_reader.cc
@@ -10,6 +10,7 @@
 #include <algorithm>
 #include <fstream>
 
+#include "reduct/core/common.h"
 #include "reduct/core/logger.h"
 
 namespace reduct::storage::io {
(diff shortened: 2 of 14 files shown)


unit_tests/reduct/api/entry_api_test.cc
```diff
@@ -131,6 +132,30 @@ TEST_CASE("EntryAPI::Write should write data with labels", "[api]") {
...
+    REQUIRE(err == Error::kOk);
+
+    auto [resp, resp_err] = receiver(R"({"name": "random"})", true);
+    REQUIRE(resp_err == Error::kOk);
+
+    auto entry = storage->GetBucket("bucket").result.lock()->GetOrCreateEntry("entry-1").result.lock();
+  }
+
+  SECTION("check headers after reading") {
+    auto [receiver, err] = EntryApi::Read(storage.get(), "bucket", "entry-1", "1000001", {});
```
atimin: it doesn't work because Catch's sections are independent. You have two runs here, one writes a record, and another one reads it. This code was called twice for each section and each one has its own unique data directory:

```cpp
  auto storage = IStorage::Build({.data_path = BuildTmpDirectory()});
  REQUIRE(storage->CreateBucket("bucket", {}) == Error::kOk);
```

Just remove the sections, you don't need them here.

src/reduct/proto/storage/entry.proto
```diff
@@ -28,6 +28,7 @@ message Record {
   uint64 end = 3;                           // end position of a blob in a block
   State state = 4;                          // state of record
   repeated Label labels = 5;                // labels
+  optional string content_type = 6;         // Content type of the record
```
atimin: you don"t need it.All fields in Protobuf 3 are optional.

src/reduct/storage/io/async_reader.cc
```diff
@@ -30,6 +31,10 @@ class AsyncReader : public async::IAsyncReader {
     for (const auto& label : record.labels()) {
       labels_.insert({label.name(), label.value()});
     }
+    content_type_ = record.content_type();
+    if (content_type_ == "") {
```
atimin: if (content_type.empty()) - more idiomatic

atimin (changes requested):

atimin (approved):
Perfect!

PR #291, opened 2023-06-06
Add header with API version

### LINKED ISSUE #249: Feature: HTTP API version in headers
### Current problem

We increase the functionality of the database and its SDKs with each release. A user might have  unaligned an SDK and a ReductStore instance and try to use a feature of the SDK that the outdated instance doesn't support.  The database provides its version with `GET /api/v1/info` method but it is inefficient to call itbefore each SDK call. 

### Possible solution

We can send `x-reduct-api` header with the current version in an HTTP response.

### Additional context

_No response_

### Code of Conduct

renghen:
Closes #249 

- [x] CHANGELOG.md have been updated (for bug fixes / features / docs)

Feature

See #249

### What is the new behavior?

The 'x-reduct-api' header is added with the API version for each response

No

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -13,11 +13,13 @@ use std::net::{IpAddr, SocketAddr};
 use std::path::PathBuf;
 
 use axum::{
+    http::header::{HeaderMap, HeaderValue},
     http::StatusCode,
     middleware,
     routing::{delete, get, head, post, put},
     Router,
 };
+use tower_default_headers::DefaultHeadersLayer;
 
 use axum_server::tls_rustls::RustlsConfig;
 
@@ -93,7 +95,10 @@ async fn main() {
         port as u16,
     );
 
-    let app = Router::new()
+    let mut headers = HeaderMap::new();
+    headers.insert("x-reduct-api", HeaderValue::from_static(version));
+
+    let app = Router::new()       
         // Server API
         .route(
             &format!("{}api/v1/info", api_base_path),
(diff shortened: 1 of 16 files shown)


src/main.rs
```diff
@@ -93,7 +95,10 @@ async fn main() {
         port as u16,
     );
 
-    let app = Router::new()
+    let mut headers = HeaderMap::new();
```
atimin: The implementation is correct but I've already created a layer `default_headers` . Could you add it there? I think it is better not to do such low level stuff in main function and keep in the `middleware` module.
Alternatively, use tower_default_headers::DefaultHeadersLayer but move the content of `default_headers` there as well, but in the `middleware.rs`

atimin (changes requested):
@rengehn , thanks for your contribution! See my comment about the implementation.

Don't forget, please, to add a record into CHANGELOG.md and a test into `api_tests/http_test.py`.

src/main.rs
```diff
@@ -93,7 +95,10 @@ async fn main() {
...
+    headers.insert("x-reduct-api", HeaderValue::from_static(version));
```
atimin: I also think that it is better to use only <major_version>.<minor_version> for the API version, because we don't change it in patch version.

atimin (changes requested):

src/main.rs
renghen: I added it in default_headers function

atimin (approved):

atimin:
@renghen , the code looks good for me. I resolved conflicts with the `main` branch and ready to merge, but the linter fails. Could you make the checks green? Use `cargo fmt`  to fix the format automatically. Thanks!

atimin:
Just figured out that I need it urgently. I've taken it with the broken CI but I'll fix it in the `main` branch.

renghen:
I did a **cargo fmt** before any push, I need to check what goes wrong on that one

renghen:
I got it, I should use 
cargo fmt --all for it to be ok. 
cargo fmt is not enough
I  did not know that.

PR #533, opened 2024-08-10
RS-359: Introduce block index file and WAL

atimin:
Closes #

Feature

### What was changed?

The PR introduces the block index file to speed up database startup. Because the block descriptors and index files are updated at different times, the RP implements WALs to store all changes until the block descriptor and index file are updated. If the database crashes, it restores the index from the WALs.

### Related issues

No

No


reductstore/src/api/bucket/get.rs
```diff
@@ -19,9 +19,9 @@ pub(crate) async fn get_bucket(
     check_permissions(&components, headers, AuthenticatedPolicy {}).await?;
     let bucket_info = components
         .storage
-        .read()
+        .write()
```
AnthonyCvn, line 22: looks like a big change from read -> to write. I'm sure you have a reason, but just to double check

reductstore/src/storage/file_cache.rs
```diff
@@ -12,7 +12,17 @@ use tokio::time::Instant;
...
+
+const FILE_CACHE_MAX_SIZE: usize = 1024;
+const FILE_CACHE_TIME_TO_LIVE: Duration = Duration::from_secs(60);
+
+pub(super) fn get_global_file_cache() -> &'static FileCache {
+    static mut FILE_CACHE: OnceLock<FileCache> = OnceLock::new();
+    unsafe {
+        FILE_CACHE.get_or_init(|| FileCache::new(FILE_CACHE_MAX_SIZE, FILE_CACHE_TIME_TO_LIVE))
+    }
+}
```
AnthonyCvn, line 25: is it at risk? By trying to search for an alternative I saw some people using lazylock: https://doc.rust-lang.org/stable/std/sync/struct.LazyLock.html#

reductstore/src/storage/block_manager/block.rs
```diff
@@ -0,0 +1,212 @@
...
+
+    pub fn insert_or_update_record(&mut self, record: Record) {
+        match self
+            .record_index
+            .entry(ts_to_us(&record.timestamp.unwrap()))
+        {
+            Entry::Occupied(mut entry) => {
+                let existing_record = entry.get_mut();
+                self.size -= existing_record.end - existing_record.begin;
+                self.metadata_size -= min(self.metadata_size, existing_record.encoded_len() as u64);
```
AnthonyCvn, line 76: does it need a warning if metadata_size is smaller than existing_record.encoded_len()?

AnthonyCvn (commented):
Was nice to review & learn at the same time 🥇  
I have noticed some codes that may need your attention.

reductstore/src/api/bucket/get.rs
atimin: Yes, this is because of the block cache. Even for read operation we change a block in cache, that is why we need to write here. I have an idea how to do this better, but I'd like to do it in the next release. Too much refactoring already.

reductstore/src/storage/file_cache.rs
atimin: It is temporary. I hate the idea of global objects, but I need to refactor many classes to avoid this singleton.

reductstore/src/storage/block_manager/block.rs
atimin: This is somehow normal. The `self.metada_size` is updated with the real size of the block descriptor when it is serialized to disk. I'm counting a rough estimate here, because I don't want to screw up the performance by serializing the descriptor for every new record to know the real size.

AnthonyCvn (approved):

PR #1186, opened 2026-02-23
Implement system  entries for attachments

### LINKED ISSUE #1019: Feature:  Support for Entry Attachments to store metadata or related artifacts.
### Current problem

An entry only has a name. However, it would be advantageous in many cases to have a description of the stored data, such as schemas and AI prompts. 

### Possible solution

Each entry may have an attachment in JSON format containing custom information. The HTTP API must provide the necessary endpoints to manage these files. It is also important that the extension has access to the attachments and can use them for data processing.

### Additional context

_No response_

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

atimin:
Closes #1019

Feature enhancement + bug fixes for attachment metadata handling via system entries.

### What was changed?

- Added system entry model for attachments under `<entry>/$meta` with dedicated behavior.
- Introduced strategy-based system behavior (`system.rs`) to centralize:
  - system entry detection,
  - write validation (`key` required),
  - upsert-by-key logic,
  - `remove=true` update handling for physical deletion,
  - rejection of `remove=true` in write path (`422`).
- Enabled hierarchical entry paths for API and storage recovery flow used by metadata entries.
- Updated bucket/entry logic to treat system entries specially for listing/count/FIFO behavior while still accounting size.
- Extended ext repository integration to read plugin config from `$meta` records.
- Added integration tests for CRUD on deep entry paths and API visibility of `$meta`.
- Added replication CI/data-check coverage for metadata replication scenarios.
- Added unit tests in `storage/entry/system.rs` for system behavior.

### Related issues

- https://github.com/reductstore/reductstore/issues/1019

No.

### Other information:

Key checks run locally:
- `cargo check -p reductstore`
- `cargo test -p reductstore storage::entry::system::tests::meta_prepare_write_removes_existing_records_with_same_key -- --nocapture`
- `cargo test -p reductstore storage::bucket::update_records::tests::remove_true_deletes_meta_record -- --nocapture`
- `cargo test -p reductstore storage::bucket::tests::get_or_create::test_get_or_create_meta_entry -- --nocapture`
- `cargo test -p reductstore storage::bucket::quotas::tests::test_fifo_quota_keeping -- --nocapture`

### DIFF
--- a/integration_tests/api/entry_api/entry_test.py
+++ b/integration_tests/api/entry_api/entry_test.py
@@ -331,9 +331,11 @@ def test_meta_entry_visibility_and_bucket_count(base_url, session, bucket):
     resp = session.get(f"{base_url}/b/{bucket}")
     assert resp.status_code == 200
     data = resp.json()
-    assert data["info"]["entry_count"] == 1
+    assert data["info"]["entry_count"] == 3
     assert data["info"]["size"] >= len(b"data") + len(b'{"v":1}')
-    assert all(entry_info["name"] != meta_entry for entry_info in data["entries"])
+    names = {entry_info["name"] for entry_info in data["entries"]}
+    assert names == {entry, "a", "a/x"}
+    assert meta_entry not in names
 
 
 def test_meta_entry_requires_key_label(base_url, session, bucket):


reductstore/src/storage/entry/system.rs
```diff
@@ -0,0 +1,337 @@
...
+///   records with the same key before adding the new one.
+/// - Updating a `$meta` record with label `remove=true` creates a tombstone (label-only update)
+///   and does not physically delete the record.
+pub(crate) const META_ENTRY_SEGMENT: &str = "$meta";
+
+/// Max block size for system metadata entries.
+pub(crate) const META_ENTRY_MAX_BLOCK_SIZE: u64 = 128 * 1024;
+
+pub(crate) fn is_system_meta_entry(entry_name: &str) -> bool {
+    entry_name == META_ENTRY_SEGMENT || entry_name.ends_with(&format!("/{}", META_ENTRY_SEGMENT))
```
AnthonyCvn, line 30: Is a meta entry at the root possible ?

reductstore/src/storage/entry/system.rs
```diff
@@ -0,0 +1,337 @@
...
+                "System entry '{}' does not support writing records with label 'remove=true'; use record update",
+                entry.name()
+            ));
+        }
+
+        let _ = entry
+            .query_remove_records(QueryEntry {
+                start: Some(0),
+                stop: Some(u64::MAX),
+                include: Some(HashMap::from([("key".to_string(), key.to_string())])),
```
AnthonyCvn, line 108: So this makes it work like a key-value store and removes the history of attachement. All good,  but just wondering if it can give any problem with replication tasks (or create some race conditions).

AnthonyCvn, line 108: to check if want to normalise entries. Because now different entries will be treated differently (/robot/temperature, and /robot/Temperature can create two different $meta entries)

atimin: Yes, it's tricky.  When the remote server receives the new update, it removes the olde records on its side. The replication happens logically.

reductstore/src/storage/entry/system.rs
atimin: Possible, not sure if  I should allow it or not.  However, it can be used as a bucket meta in the future.

reductstore/src/storage/entry/system.rs
atimin: yes, but the storage was always case-sensitive. It's a huge compatibility break to change it.

PR #1238, opened 2026-03-24
Audit bucket

### LINKED ISSUE #1170: Feature: Audit Bucket
Audit bucket which saves requests per endpoint.

- Count Method
- Duration
- Aggregation Methods

tuanhungngyn:
Closes #1170

- [ ] Tests for the changes have been added (for bug fixes / features)

Feature

### What was changed?

This PR introduces audit logging backed by a reserved system bucket, `$audit`.

Key changes:
- Add audit event logging from HTTP middleware.
- Store audit events as JSON records in `$audit`.
- Aggregate repeated requests before flush using a worker + channel pattern with a 5-second inactivity window.
- Treat `$audit` like a normal bucket for read/write APIs, but enforce stricter token matching for all system buckets (`$...`).
- Wildcard and prefix permissions do not grant access to system buckets; exact match is required.
- `full_access = true` still bypasses bucket checks and allows admin access.
- Add internal `create_system_bucket()` so system buckets can be created internally without exposing that through public bucket creation rules.
- In replica mode, forward `log_event` writes over HTTP with primary/secondary failover and last-successful-URL caching.

### Related issues

- #1170

Behavior change for token permissions:
- Buckets whose names start with `$` are now treated as system buckets.
- Wildcard or prefix permissions no longer grant access to system buckets.
- Non-admin tokens must have an exact permission match such as `$audit` to access them.

New environment variables for replica audit forwarding:
- `RS_PRIMARY_URL`
- `RS_SECONDARY_URL`

### Other information:

Current implementation direction is based on the updated design decision to avoid a dedicated audit endpoint and instead rely on normal bucket APIs with stricter system-bucket permission rules.

### DIFF
--- a/reductstore/src/api/http/middleware.rs
+++ b/reductstore/src/api/http/middleware.rs
@@ -217,7 +217,7 @@ fn log_level_for_response(status: StatusCode, skip_error_log: bool) -> Level {
 mod tests {
     use super::*;
     use crate::api::components::StateKeeper;
-    use crate::api::http::tests::{api_limited_keeper, keeper};
+    use crate::api::http::tests::{api_limited_keeper, keeper, waiting_keeper};
     use crate::audit::{AuditEvent, AUDIT_BUCKET_NAME};
     use axum::http::Request;
     use axum::http::{HeaderMap, HeaderValue};
@@ -255,6 +255,15 @@ mod tests {
         assert!(!headers
             .get("x-reduct-log-hint")
             .is_some_and(|v| v == "skip-error-log"));
+    }
+
+    #[rstest]
+    #[case("/alive", true)]
+    #[case("/ready", true)]
+    #[case("/api/v1/audit/token", true)]
+    #[case("/api/v1/info", false)]
+    fn checks_audit_skip_paths(#[case] path: &str, #[case] expected: bool) {
+        assert_eq!(should_skip_audit(path), expected);
     }
 
     #[rstest]
@@ -369,6 +378,40 @@ mod tests {
         assert_eq!(event.endpoint, "GET /protected");
         assert_eq!(event.status, StatusCode::UNAUTHORIZED.as_u16());
         assert_eq!(event.call_count, 1);
+    }
+
+    #[rstest]
+    #[tokio::test]
+    async fn resolves_token_name_for_authenticated_request(#[future] keeper: Arc<StateKeeper>) {
+        let keeper = keeper.await;
+        let components = keeper.get_anonymous().await.unwrap();
+
+        let token_name =
+            resolve_audit_token_name(StatusCode::OK, Some("Bearer init-token"), Some(&components))
+                .await;
+
+        assert_eq!(token_name, Some("init-token".to_string()));
+    }
+
+    #[rstest]
+    #[tokio::test]
+    async fn returns_none_for_missing_auth_header(#[future] keeper: Arc<StateKeeper>) {
+        let keeper = keeper.await;
+        let components = keeper.get_anonymous().await.unwrap();
+
+        let token_name = resolve_audit_token_name(StatusCode::OK, None, Some(&components)).await;
+
+        assert_eq!(token_name, None);
+    }
+
+    #[rstest]
+    #[tokio::test]
+    async fn returns_none_when_components_are_unavailable(
+        #[future] waiting_keeper: Arc<StateKeeper>,
+    ) {
+        let keeper = waiting_keeper.await;
+        let components = get_audit_components(&keeper).await;
+        assert!(components.is_none());
     }
 
     #[rstest]
--- a/reductstore/src/audit.rs
+++ b/reductstore/src/audit.rs
@@ -50,3 +50,83 @@ impl AuditRepositoryBuilder {
 }
 
 pub(crate) type BoxedAuditRepository = Box<dyn ManageAudit + Send + Sync>;
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::audit::aggregator::AGGREGATION_WINDOW_SECS;
+    use crate::cfg::Cfg;
+    use reduct_base::io::ReadRecord;
+    use rstest::{fixture, rstest};
+    use tempfile::tempdir;
+    use tokio::time::{sleep, Duration};
+
+    #[fixture]
+    async fn storage_and_cfg() -> (Arc<StorageEngine>, Cfg) {
+        let tmp_dir = tempdir().unwrap();
+        let cfg = Cfg {
+            data_path: tmp_dir.keep(),
+            ..Cfg::default()
+        };
+        let storage = Arc::new(
+            StorageEngine::builder()
+                .with_data_path(cfg.data_path.clone())
+                .with_cfg(cfg.clone())
+                .build()
+                .await,
+        );
+        (storage, cfg)
+    }
+
+    fn make_event() -> AuditEvent {
+        AuditEvent {
+            timestamp: 1,
+            token_name: "token-1".to_string(),
+            endpoint: "GET /api/v1/info".to_string(),
+            status: 200,
+            call_count: 1,
+            duration: 100,
+        }
+    }
+
+    #[rstest]
+    #[tokio::test(flavor = "multi_thread")]
+    async fn builder_uses_local_repository_for_non_replica(
+        #[future] storage_and_cfg: (Arc<StorageEngine>, Cfg),
+    ) {
+        let (storage, cfg) = storage_and_cfg.await;
+        let mut repo = AuditRepositoryBuilder::new(cfg)
+            .build(Arc::clone(&storage))
+            .await;
+
+        repo.log_event(make_event()).await.unwrap();
+        sleep(Duration::from_millis(AGGREGATION_WINDOW_SECS * 1000 + 300)).await;
+
+        let bucket = storage
+            .get_bucket(AUDIT_BUCKET_NAME)
+            .await
+            .unwrap()
+            .upgrade_and_unwrap();
+        let mut reader = bucket.begin_read("token-1", 1).await.unwrap();
+        let record = reader.read_chunk().unwrap().unwrap();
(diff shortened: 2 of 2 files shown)


reductstore/src/api/http/utils.rs
```diff
@@ -22,6 +23,20 @@ use std::sync::Arc;
 use std::task::{Context, Poll};
 use tokio::sync::Mutex;
 
+pub(super) fn ensure_public_bucket(bucket_name: &str) -> Result<(), HttpError> {
```
atimin: it's available, we don't need to check. We only need a special policy for tokens that the wildcards don't reveal system bucket.

reductstore/src/audit/repo.rs
```diff
@@ -0,0 +1,391 @@
...
+}
+
+#[async_trait]
+impl ManageAudit for AuditRepository {
+    async fn log_event(&mut self, event: AuditEvent) -> Result<(), ReductError> {
+        let (key, version) = self.aggregate_event(event).await?;
+        let state = Arc::clone(&self.state);
+        let storage = Arc::clone(&self.storage);
+        tokio::spawn(async move {
+            AuditRepository::flush_after_delay(state, storage, key, version).await;
```
atimin: i think we can't spin up a task for every evernt. It must  worker + channel pattern.


in constructor:

```rust
{
   let (tx, rx) = channel(SIZE);

  tokio:spawn(async move {
        while let Some(even( = rx.recv.await {

       }  

  })

}
```


```rust
 async fn log_event(&mut self, event: AuditEvent) -> Result<(), ReductError> {
   self.tx.send(event).await;
 }
```

reductstore/src/audit/read_only.rs
```diff
@@ -0,0 +1,166 @@
...
+pub(crate) struct ReadOnlyAuditRepository {
+    client: Client,
+    primary_url: String,
+}
+
+impl ReadOnlyAuditRepository {
+    pub async fn new(cfg: Cfg, _storage: Arc<StorageEngine>) -> Self {
+        let primary_url = cfg
+            .replication_conf
+            .primary_url
```
atimin: I thought about it a bit more. I think primary alone isn't enough. We need to add the secondary URL.


The logic is as follows:

1. Try primary.
2. If that fails, write to the secondary.
3. Remember the last successful URL to avoid writing to an unavailable instance.

reductstore/src/storage/engine.rs
```diff
@@ -317,8 +318,9 @@ impl StorageEngine {
         let infos = {
             let buckets = self.buckets.read().await?;
             buckets
-                .values()
-                .map(|bucket| bucket.clone().info())
+                .iter()
+                .filter(|(name, _)| name.as_str() != AUDIT_BUCKET_NAME)
```
atimin: If I have access to it, I should be able to see it; don't filter. Admins, for example, should have access by default.

reductstore/src/storage/engine.rs
```diff
@@ -372,6 +374,9 @@ impl StorageEngine {
 
 pub(super) fn check_name_convention(name: &str) -> Result<(), ReductError> {
     let regex = regex::Regex::new(r"^[A-Za-z0-9_-]*$").unwrap();
+    if name == AUDIT_BUCKET_NAME {
```
atimin: create a method to create_system_bucket(), which isn't available through the public APIs. There you can ignore rules.

reductstore/src/api/http/audit.rs
```diff
@@ -0,0 +1,55 @@
...
+    start: Option<u64>,
+    end: Option<u64>,
+    endpoint: Option<String>,
+}
+
+pub(crate) fn create_audit_api_routes() -> Router<Arc<StateKeeper>> {
+    Router::new().route("/{token_name}", get(get_audit_events))
+}
+
+pub(super) async fn get_audit_events(
```
atimin: We don't need a special endpoint; admins and users with an audit token can access it. 

Remove query_token_events from the trait

reductstore/src/audit/read_only.rs
```diff
@@ -0,0 +1,166 @@
...
+        .and_then(|v| v.to_str().ok())
+        .unwrap_or("Unknown");
+
+    Err(ReductError::new(status, error_msg))
+}
+
+#[async_trait]
+impl ManageAudit for ReadOnlyAuditRepository {
+    async fn log_event(&mut self, _event: AuditEvent) -> Result<(), ReductError> {
+        Err(forbidden!("Cannot write audit data in read-only mode"))
```
atimin: you should.

reductstore/src/audit/repo.rs
```diff
@@ -0,0 +1,388 @@
...
+    ) -> Result<(), ReductError> {
+        let bucket = match storage.get_bucket(AUDIT_BUCKET_NAME).await {
+            Ok(bucket) => bucket.upgrade()?,
+            Err(_) => storage
+                .create_bucket(AUDIT_BUCKET_NAME, Bucket::defaults())
+                .await?
+                .upgrade()?,
+        };
+
+        let labels = Labels::from([("endpoint".to_string(), event.endpoint.clone())]);
```
atimin: I would label only status. endpoint isn't that important.

reductstore/src/cfg/replication.rs
```diff
@@ -19,6 +19,8 @@ pub struct ReplicationConfig {
     pub verify_ssl: bool,
     /// Optional custom CA certificate path for replication destinations
     pub ca_path: Option<PathBuf>,
+    /// Primary service URL for replica-side internal HTTP fetches
+    pub primary_url: Option<String>,
```
atimin: It's in the wrong place; it's the other replications that stream data. It should be where the RS_INSTANCE_ROLE is set.

reductstore/src/auth/policy.rs
```diff
@@ -113,11 +115,17 @@ impl Policy for WriteAccessPolicy<'_> {
 
 fn check_bucket_permissions(token_list: &Vec<String>, bucket: &str) -> bool {
     for token_bucket in token_list {
+        if token_bucket == bucket {
+            return true;
+        }
+
+        if bucket == AUDIT_BUCKET_NAME {
```
atimin: I would make it generic. Anything that starts with a "$" is a system bucket and the token must be an exact match. 

Set full_access to true to bypass all checks and give admins access to all system buckets.

reductstore/src/audit/repo.rs
atimin: Check the 'ReplicationSender' and 'LockFile' examples.

reductstore/src/audit/repo.rs
```diff
@@ -0,0 +1,439 @@
...
+        };
+
+        for event in events {
+            Self::write_event_to_bucket(Arc::clone(storage), event).await?;
+        }
+
+        Ok(())
+    }
+
+    fn into_event(key: AuditAggregateKey, aggregate: AuditAggregate) -> AuditEvent {
```
atimin: Ambiguous name, "into" is always self-moving operation in Rust. call it make_event or something like this.

reductstore/src/audit/repo.rs
```diff
@@ -0,0 +1,439 @@
...
+    }
+}
+
+#[async_trait]
+impl ManageAudit for AuditRepository {
+    async fn log_event(&mut self, event: AuditEvent) -> Result<(), ReductError> {
+        self.tx
+            .send(event)
+            .await
+            .map_err(|_| ReductError::internal_server_error("Audit worker is not available"))
```
atimin: internal_server_error! - use macros.

reductstore/src/audit/repo.rs
```diff
@@ -0,0 +1,439 @@
...
+        }
+    }
+
+    async fn write_event_to_bucket(
+        storage: Arc<StorageEngine>,
+        event: AuditEvent,
+    ) -> Result<(), ReductError> {
+        let labels = Labels::from([("status".to_string(), event.status.to_string())]);
+        let payload = serde_json::to_vec(&event).map_err(|err| {
+            ReductError::internal_server_error(&format!("Failed to serialize audit event: {}", err))
```
atimin: internal_server_error!

reductstore/src/audit/read_only.rs
```diff
@@ -0,0 +1,224 @@
...
+        if !cfg.api_token.is_empty() {
+            let mut value = HeaderValue::from_str(&format!("Bearer {}", cfg.api_token)).unwrap();
+            value.set_sensitive(true);
+            headers.insert(reqwest::header::AUTHORIZATION, value);
+        }
+
+        let mut builder = Client::builder()
+            .default_headers(headers)
+            .connect_timeout(cfg.replication_conf.connection_timeout)
+            .danger_accept_invalid_certs(!cfg.replication_conf.verify_ssl)
```
atimin: it's also not right to use settings from another feature. let's hand-code for now. accept invalid certs, timeout is 5s.

reductstore/src/audit/read_only.rs
```diff
@@ -0,0 +1,224 @@
...
+            builder = builder.add_root_certificate(cert);
+        }
+
+        builder.build().map_err(|err| {
+            unprocessable_entity!("Failed to build audit replica HTTP client: {}", err)
+        })
+    }
+
+    async fn log_event_with_failover(&self, event: &AuditEvent) -> Result<(), ReductError> {
+        let mut candidates = Vec::new();
```
atimin, line 79: didn't find aggregation.

reductstore/src/audit/read_only.rs
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    async fn log_event_to_url(
+        &self,
+        base_url: &str,
+        event: &AuditEvent,
+    ) -> Result<(), ReductError> {
+        let url = build_write_url(base_url, event);
+        let payload = serde_json::to_vec(event).map_err(|err| {
+            ReductError::internal_server_error(&format!("Failed to serialize audit event: {}", err))
```
atimin: macros

PR #1245, opened 2026-03-25
Fix replication mode reprovisioning to override only on explicit MODE env

mother-6000:
Closes #

Bug fix

### What was changed?

- fixed replication provisioning so existing replication mode is preserved by default during config reprovisioning
- added explicit mode-override tracking while parsing env vars (`RS_REPLICATION_<id>_MODE`)
- changed conflict-update logic to override persisted mode only when the MODE env var was explicitly provided
- added a regression test to verify that an existing replication mode is overridden when `RS_REPLICATION_1_MODE` is set

### Related issues

N/A

No

### Other information:

Could not run tests in this environment because the installed Cargo toolchain is too old for the workspace lockfile/crate editions (requires newer nightly/edition2024 support).

### DIFF
--- a/reductstore/src/cfg.rs
+++ b/reductstore/src/cfg.rs
@@ -39,7 +39,7 @@ use reduct_base::msg::bucket_api::BucketSettings;
 use reduct_base::msg::replication_api::ReplicationSettings;
 use reduct_base::msg::server_api::License;
 use reduct_base::msg::token_api::Token;
-use std::collections::{HashMap, HashSet};
+use std::collections::HashMap;
 use std::fmt::{Display, Formatter};
 use std::path::PathBuf;
 use std::sync::Arc;
@@ -61,6 +61,12 @@ pub enum InstanceRole {
     Replica,
 }
 
+#[derive(Clone, Default)]
+pub struct ProvisionedReplication {
+    pub settings: ReplicationSettings,
+    pub mode_override: Option<String>,
+}
+
 #[derive(Clone)]
 pub struct Cfg {
     pub log_level: String,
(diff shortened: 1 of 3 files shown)


reductstore/src/cfg.rs
```diff
@@ -281,6 +284,7 @@ impl<EnvGetter: GetEnv, ExtCfg: ExtCfgBounds> CfgParser<EnvGetter, ExtCfg> {
             env,
             license,
             ext_cfg,
+            replications_with_mode_override,
```
atimin: i don't like it, let's change replications field:

it will have the following type: 
```rust
struct ProvisionedReplication {
   settings: ReplicationSettings,
   mode_override:  Option<String>
}

Cfg {
..
    pub replications: HashMap<String, ProvisionedReplication>
   
}
```

reductstore/src/cfg.rs
```diff
@@ -163,6 +163,7 @@ pub struct CfgParser<EnvGetter: GetEnv = StdEnvGetter, ExtCfg: ExtCfgBounds = Co
     pub license: Option<License>,
     pub env: Env<EnvGetter>,
     pub ext_cfg: ExtCfg,
+    pub replications_with_mode_override: HashSet<String>,
```
atimin: remove

atimin (changes requested):
update changelog

reductstore/src/cfg.rs
mother-6000: Done in c44bb39. I refactored `Cfg.replications` to `HashMap<String, ProvisionedReplication>` and moved `mode_override` into the struct (`reductstore/src/cfg.rs`, `reductstore/src/cfg/provision/replication.rs`). Provisioning logic now preserves persisted mode when `mode_override` is not set, and overrides only when `RS_REPLICATION_<ID>_MODE` is explicitly provided.

reductstore/src/cfg.rs
mother-6000: Removed in c44bb39 along with the requested refactor in `cfg.rs`.

PR #1286, opened 2026-04-03
Fix /alive and /ready returning 503 during lock wait

### LINKED ISSUE #1285: /alive returns 503 during lock wait (should be 200)
`/alive` can return `503` during lock-file wait, but it should return `200`.

Cause: global `check_api_rate_limit` middleware runs before route handlers and calls `get_anonymous()`. During lock wait that returns startup `503`, so `/alive` handler never gets a chance to map it to `200`.

Expected:
- `GET /api/v1/alive` -> `200` even while waiting for lock.

Fix:
- In `check_api_rate_limit`, skip `/alive` and `/ready` before calling `get_anonymous()`.

mother-6000:
## Summary
- skip API rate-limit middleware for /alive and /ready paths before calling get_anonymous()
- allow both direct (/alive, /ready) and prefixed (/api/v1/alive, /api/v1/ready) paths
- add middleware tests covering skipped health endpoints under waiting state

## Testing
- cargo test -p reductstore skips_api_rate_limit_for_health_endpoints
- cargo test -p reductstore enforces_api_rate_limit

Closes #1285

### DIFF
--- a/reductstore/src/api/http/middleware.rs
+++ b/reductstore/src/api/http/middleware.rs
@@ -62,20 +62,23 @@ pub(super) async fn check_api_rate_limit(
     request: Request<Body>,
     next: Next,
 ) -> Result<impl IntoResponse, HttpError> {
-    if should_skip_api_rate_limit(request.uri().path()) {
-        return Ok(next.run(request).await);
-    }
+    let path = request.uri().path();
 
-    let components = keeper.get_anonymous().await?;
+    let components = match keeper.get_anonymous().await {
+        Ok(components) => components,
+        Err(err)
+            if err.status() == ErrorCode::ServiceUnavailable && path.ends_with("/api/v1/alive") =>
+        {
+            return Ok(next.run(request).await);
+        }
+        Err(err) => return Err(err.into()),
+    };
+
     if let Err(err) = components.limits.check_api_request().await {
         return Err(HttpError::from(err));
     }
 
     Ok(next.run(request).await)
-}
-
-fn should_skip_api_rate_limit(path: &str) -> bool {
-    path.ends_with("/alive") || path.ends_with("/ready")
 }
 
 pub async fn print_statuses(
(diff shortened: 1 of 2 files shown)


reductstore/src/api/http/middleware.rs
```diff
@@ -70,6 +74,10 @@ pub(super) async fn check_api_rate_limit(
     Ok(next.run(request).await)
 }
 
+fn should_skip_api_rate_limit(path: &str) -> bool {
+    path.ends_with("/alive") || path.ends_with("/ready")
```
atimin: use the full path. it can be an entry name.

reductstore/src/api/http/middleware.rs
```diff
@@ -62,6 +62,10 @@ pub(super) async fn check_api_rate_limit(
     request: Request<Body>,
     next: Next,
 ) -> Result<impl IntoResponse, HttpError> {
+    if should_skip_api_rate_limit(request.uri().path()) {
```
atimin: I've changed my mind. Let's check limits for /alive and /ready.  but for alive   we map 512 -> 200 like we did it in the /alive endpoint.

mother-6000: Done. I reverted the skip approach: limits are now still checked for `/ready`, and for `/api/v1/alive` only we map startup `503` from `get_anonymous()` to continue into the endpoint flow (which returns `200`). Added focused tests for alive/ready behavior in waiting state.

reductstore/src/api/http/middleware.rs
mother-6000: Good catch — updated to match the full API path (`/api/v1/alive`) instead of a generic suffix, and added a regression test for `/api/v1/b/bucket/entry/alive` to ensure it does not trigger the alive exception.

atimin (approved):

PR #1376, opened 2026-05-08
add lifecycle policy background workers

### LINKED ISSUE #1012: Background Worker: Enforce Retention Policies (+ ENV provisioning)
### Summary
Implement a **background task** that enforces bucket-scoped **retention policies** (created via HTTP API) by deleting matching **records**.  
Also support **provisioning via environment variables** (mirrors replication provisioning style).

### ENV Provisioning (example)

```
RS_RETENTION_A_NAME=purge-sensors-30d
RS_RETENTION_A_BUCKET=telemetry
RS_RETENTION_A_ENTRIES=sensors/**,env/{temp,humidity}
RS_RETENTION_A_MAX_AGE=30d
RS_RETENTION_A_WHEN={ "&label":  {"$eq": true}
```

atimin:
Closes #1012

Feature

### What was changed?

- Added lifecycle policy configuration and environment provisioning for delete lifecycles.
- Added lifecycle repository/task infrastructure to persist lifecycle settings, start/stop background workers, and execute delete actions on configured intervals.
- Integrated lifecycle runs with audit logging using structured audit event payloads shared by API and lifecycle events.
- Updated audit logger structure and tests to support typed events and payload-based assertions.
- Adjusted CI workflow runner configuration for this branch.

### Related issues

- #1012

No user-facing API break is expected. Audit records are now stored as structured events with a `payload` object, so consumers that read raw `$audit` records directly may need to account for the nested payload format.

### Other information:

Validation performed in commits on this branch includes Rust unit-test updates for lifecycle provisioning, repository behavior, worker startup/shutdown, and audit middleware assertions.


reduct_base/src/msg/lifecycle_api.rs
```diff
@@ -0,0 +1,169 @@
...
+    /// Lifecycle policy action type.
+    #[serde(default, rename = "type")]
+    pub lifecycle_type: LifecycleType,
+    /// Bucket to apply the lifecycle policy to.
+    pub bucket: String,
+    /// Entries to clean. If empty, all removable entries are matched. Prefix wildcards are supported.
+    #[serde(default)]
+    pub entries: Vec<String>,
+    /// Maximum record age, e.g. "30d", "24h", or "3600s".
+    pub max_age: String,
```
AnthonyCvn, line 48: would make sense to reject max_age below `1h`

AnthonyCvn (changes requested):

reduct_base/src/msg/lifecycle_api.rs
```diff
@@ -0,0 +1,169 @@
...
+    Delete,
+}
+
+/// Lifecycle mode.
+#[repr(u8)]
+#[derive(Serialize, Deserialize, Clone, Copy, Debug, PartialEq, Eq)]
+#[serde(rename_all = "lowercase")]
+pub enum LifecycleMode {
+    /// Lifecycle is active and executes actions.
+    Enabled,
```
atimin, line 22: Should the query data be removed or converted to another type?  Null action.

reduct_base/src/msg/lifecycle_api.rs
atimin: Maybe not even on the same day. I also think it should not be run immediately, but after the interval has expired.  So a user could have time to remove it if it's wrong.

reduct_base/src/msg/lifecycle_api.rs
AnthonyCvn: probably useless to have dryrun; the idea was only to make sure that the max_age is right. But we can also Disable, set the parameter and check. Also if max_age has a minimum allowed; then it's not a problem.

reduct_base/src/msg/lifecycle_api.rs
AnthonyCvn: influx has it constrained to 1h. A min interval would be great you're right

AnthonyCvn (approved):

PR #1387, opened 2026-05-20
Optimize remote storage synchronization for primary and read-only replicas

### LINKED ISSUE #1386: Bug: S3 block sync freezes DB for several seconds when many entries have received data
## Summary

When a ReductStore instance has many entries (roughly 200–300) that have received data, the database appears to freeze for several seconds during each sync interval while trying to sync blocks to an S3 backend.

## Impact

This causes:
- CLI timeout errors
- healthcheck failures/timeouts
- temporary unresponsiveness during the sync window

## Observed behavior

- ReductStore runs normally between sync intervals.
- On each sync interval, when block sync to S3 starts, the database/store becomes frozen for several seconds.
- The freeze duration is long enough to trigger timeouts in CLI operations and health checks.

## Expected behavior

Syncing blocks to the S3 backend should not block the database/store for multiple seconds, even when the instance has a few hundred active entries.

## Environment

- Repository: `reductstore/reductstore`
- Backend involved: S3 block sync
- Number of affected entries: approximately 200–300 entries with received data

## Notes

This looks like sync work is blocking normal request handling or creating long stalls under higher entry counts.

If needed, I can provide more details about configuration, sync interval, backend, and logs.

atimin:
Closes #1386

Bug fix / performance improvement

### What was changed?

- Reworked read-only replica synchronization to run in the background instead of blocking request paths.
- Optimized file-cache and WAL synchronization flow to reduce long pauses when many entries have pending data.
- Improved block-manager and read-only bucket/engine sync behavior, including local WAL checks and reduced lock contention during compaction-related paths.
- Added resiliency improvements around transient sync errors and non-critical remove failures to avoid noisy warnings and unnecessary disruption.
- Updated related storage, write path, and query internals to align with asynchronous replica sync behavior.
- Updated dependencies (`Cargo.lock`).

### Related issues

- https://github.com/reductstore/reductstore/issues/1386

No.

### Other information:

This branch includes all commits currently on `1386-bug-s3-block-sync-freezes-db-for-several-seconds-when-many-entries-have-received-data` relative to `main`, not only the initial file-cache batching change.

### DIFF
--- a/reductstore/src/launcher.rs
+++ b/reductstore/src/launcher.rs
@@ -447,12 +447,69 @@ mod tests {
         assert!(bucket_names.contains(&"bucket-2".to_string()));
     }
 
+    #[rstest]
+    #[tokio::test(flavor = "multi_thread")]
+    #[serial]
+    async fn test_launch_spawns_replica_reload_for_replica_role_with_non_zero_interval() {
+        let reloads = Arc::new(AtomicUsize::new(0));
+        test_observer::set_replica_reload_observer(Some(reloads.clone()));
+
+        let mut cfg = HashMap::new();
+        cfg.insert("RS_INSTANCE_ROLE".to_string(), "REPLICA".to_string());
+        cfg.insert(
+            "RS_ENGINE_REPLICA_UPDATE_INTERVAL".to_string(),
+            "1".to_string(),
+        );
+
+        let task = set_env_and_run(cfg).await;
+        sleep(Duration::from_millis(300)).await;
+
+        *STOP_SERVER.lock().await = true;
+        task.join().unwrap();
+        test_observer::set_replica_reload_observer(None);
+
+        assert!(
+            reloads.load(Ordering::Relaxed) > 0,
+            "replica reload task should be spawned for replica role with non-zero interval"
+        );
+    }
+
+    #[rstest]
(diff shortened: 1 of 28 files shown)


reductstore/src/launcher.rs
```diff
@@ -181,34 +196,93 @@ mod test_observer {
...
+
+        let execution_time = started_at.elapsed();
+        if execution_time > interval {
+            warn!(
+                "Periodic {} took {:?}, exceeding configured interval {:?}",
+                task_name, execution_time, interval
+            );
         }
+
+        next_tick += interval;
```
AnthonyCvn, line 274: If `execution_time > interval` there is a warning but the `next_tick` will be in the past and execution will try to catch up. For a loop that runs forever, why not`next_tick = tokio::time::Instant::now() + interval` ? or simply `sleep(interval)` at every loop

atimin: it's intentional.  The tasks are asynchronous to the storage engine , it's fine that it's not sleeping since it switch the tokio context during IO operations. 

The idea:

1. Do the best trying to update or compact the meta information
2.  If fail, show a user warning that you can't update in set period. The user should change the settings.

AnthonyCvn (approved):

PR #1413, opened 2026-06-05
feat(storage): add read-only storage doctor for offline data-path inspection

khoinp1012:
## Add read-only storage doctor for offline data-path inspection

This PR adds a new `storage::doctor` module that scans a local filesystem ReductStore data path and reports storage consistency problems without mutating any files.

### What it does

The doctor walks the data path and reports issues for:

- Missing or unreadable data path
- Buckets without or with corrupt `bucket.settings`
- Entries discovered by `blocks.idx` (including nested entries)
- Block index CRC mismatch, empty index with descriptors, decode failure
- Index references missing descriptor or data block
- Descriptor CRC mismatch with index
- Descriptor `begin_time` mismatch with block id
- Descriptor decode failure
- Data block smaller than descriptor `size`
- Orphan descriptor and data block files
- Unrecognized files inside entries

### Why

ReductStore currently exposes WAL/block/index recovery only through server startup. Operators have no offline way to inspect storage health before bringing up the database, after a power loss, or before a backup.

### Design notes

- Strictly read-only. No `BlockIndex::try_load`, no `FolderKeeper`, no `create_wal`, no `EntryLoader::restore_*` calls.
- All protos, timestamps, and constants come from the project's existing `crate::storage::proto` and existing extension constants.
- CRC computation matches the field order used by `BlockIndex::from_proto` (`block_id`, `size`, `record_count`, `metadata_size`, `latest_record_time`, optional `descriptor_crc`).
- Empty `blocks.idx` with descriptors alongside is reported as `CorruptBlockIndex` to match `BlockIndex::try_load` behavior.
- `bucket.settings` is decoded as `proto::BucketSettings`; decode failure reports `InvalidBucketSettings`.
- Skips `.wal` and legacy `wal` directories during entry discovery.

### Tests

- 13 unit tests covering: missing data path, empty data path, nested entries, internal WAL skip, missing bucket settings, invalid bucket settings, corrupt block index, empty block index with descriptors, missing descriptor/data block, orphan descriptor/data block, descriptor CRC mismatch, data block too small, valid index with matching files.
- All `cargo test -p reductstore storage::doctor` tests pass.
- `cargo check --workspace` passes.
- `cargo fmt --all --check` passes.

### Out of scope

- Repair or apply commands.
- S3/remote backend scanning.
- Server HTTP endpoint.

### Files changed

- `reductstore/src/storage.rs`: declare `pub mod doctor;`
- `reductstore/src/storage/doctor.rs`: new module with report types, scanner, and tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add read-only storage doctor for offline data-path inspection, by @khoinp1012, [PR-1413](https://github.com/reductstore/reductstore/pull/1413)
 - Accept `x-reduct-content-length` as an alternative to `Content-Length` for browser streaming via the Fetch API, [PR-1411](https://github.com/reductstore/reductstore/pull/1411)
 - Pass attachment key-value maps to extensions instead of selecting a single attachment key, [PR-1383](https://github.com/reductstore/reductstore/pull/1383)
 - Add HTTP CRUD endpoints for lifecycle policies, [PR-1382](https://github.com/reductstore/reductstore/pull/1382)
--- a/reductstore/src/bin/reduct-doctor.rs
+++ b/reductstore/src/bin/reduct-doctor.rs
@@ -1,0 +1,433 @@
+// Copyright 2021-2026 ReductSoftware UG
+// Licensed under the Apache License, Version 2.0
+
+//! Read-only storage doctor for offline ReductStore data-path inspection.
+//!
+//! Scans a local filesystem data path and reports storage consistency
+//! problems without mutating any files. Strictly read-only: no server
+//! startup, no log forwarding, no remote backend access.
+
+use reductstore::storage::doctor::{
+    scan_data_path, DoctorIssue, DoctorReport, DoctorSeverity, DoctorSummary,
+};
+use std::path::PathBuf;
+use std::process::ExitCode;
+
+const DEFAULT_DATA_PATH: &str = "/data";
+const DEFAULT_FAIL_ON: &str = "error";
+
+fn main() -> ExitCode {
+    let args: Vec<String> = std::env::args().collect();
+
+    if args.iter().any(|arg| arg == "--version" || arg == "-V") {
+        println!("reduct-doctor {}", env!("CARGO_PKG_VERSION"));
+        return ExitCode::SUCCESS;
+    }
+
+    if args.iter().any(|arg| arg == "--help" || arg == "-h") {
+        print_help();
+        return ExitCode::SUCCESS;
+    }
+
+    let opts = match parse_args(&args[1..]) {
+        Ok(opts) => opts,
+        Err(err) => {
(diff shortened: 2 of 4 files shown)


atimin:
Hi @khoinp1012, thanks a lot for your contribution!

I like the idea and could merge it, but it would be great to make the feature available as a command. 
Let's explore the options:
1) It could be made available as the "reductstore doctor" subcommand.
2) Or it could be a separate command, 'reduct-doctor'. 

Personally, I would prefer it to be a separate binary (https://doc.rust-lang.org/cargo/reference/cargo-targets.html#binaries). 

Do you have time to elaborate it a bit more?

P.S. Please add a record to the changelog with "by @khoinp1012".

khoinp1012:
Thanks for the review! I've addressed your feedback in three new commits on top of the original PR:

1. **docs(changelog):** added an entry under `## [Unreleased] / ### Added` crediting @khoinp1012 and linking PR-1413.

2. **test(storage):** added 5 unit tests (`doctor_accepts_valid_bucket_settings`, `doctor_reports_descriptor_block_id_mismatch`, `doctor_reports_invalid_index_proto_decode`, `doctor_indexes_with_and_without_descriptor_crc`, `doctor_reports_unrecognized_entry_files`) to lift patch coverage above the Codecov threshold. Doctor tests now 18/18 pass.

3. **feat(storage):** added a separate `reduct-doctor` binary under `reductstore/src/bin/` (auto-discovered by Cargo, no Cargo.toml change). It implements `--version`, `--help`, `--data-path` (with `$RS_DATA_PATH` fallback), `--format human|json`, and `--fail-on error|warning|info`, with exit codes 0/1/2. 10 unit tests cover arg parsing and threshold logic.

I tested the binary locally:
- `--version` and `--help` exit 0
- healthy tempdir exits 0
- corrupt index exits 1 with a clear `CorruptBlockIndex` issue
- missing path exits 1 with `MissingDataPath`
- `--bogus` exits 2 with usage error
- missing value exits 2

I did not extend `.github/actions/build-binaries`, `build-docker`, or release tarball/zip assets to ship `reduct-doctor` in this PR. Happy to add a follow-up PR that updates the CI artifacts to include the new binary if you want it in the Docker image and release assets too.

Let me know if you'd like any other changes.

atimin:
Hi @khoinp1012,
I've decided not to accept your pull request. Having checked the code, I found several issues. It looks like AI-generated code that doesn't reflect how things actually work. 

While I like the general idea of building a file system health checker, the implementation requires a deep understanding of the data engine. This isn't a simple task that can be done in one step with an LLM. We may start working on it in the future, but we don't have the capacity to do so today.

Thanks for your efforts.

khoinp1012:
Hi atimin,

I think you should change the approach. LLM-generated code almost unavoidable nowadays. Create new branch, well test before merging it to main is much better approach than avoid it. Beta - fix - release, is always the basic release cycle of software.
I design this project, but AI implemented it. https://github.com/khoinp1012/kprotect
I am a system designer, who use AI to code ! I am not a vibe coder dont understand how things work.
AI will not have idea of storage doctor !

PR #1417, opened 2026-06-06
Feat/replication audit events

### LINKED ISSUE #1374: Replication: aggregated audit events for sync progress and errors
## Problem statement
Replication currently does not report aggregated operational telemetry into the audit bucket. This makes it hard to monitor replication health and progress (errors, synced records, pending records) in a consistent way.

## Proposed solution
Provide the replication module with access to the audit bucket and emit aggregated replication metrics in JSON format. Reporting cadence should be every 1 minute, or every 5 seconds when there is no activity. The event schema and exact metrics will be finalized during implementation.

DibbayajyotiRoy:
Closes #1374

Feature.

### What was changed?

Replication tasks now emit periodic, aggregated diagnostics as `$system` events,
mirroring the lifecycle system-event wiring (#1399) and reusing the audit aggregator's
timing model.

- New `$system` entry path: `replications/<instance-name>/<replication-name>`
- One active bucket per replication task, keyed by status code. Flushed on (1) status
  change, (2) 5s idle, (3) 60s cap — reusing AGGREGATION_WINDOW_SECS / FORCE_FLUSH_TIMEOUT_SECS.
- Unified payload schema for success and failure events:
  `{ status, pending_records, written_records, failed_records, replicated_data_size, duration }`
  A 200→404 transition flushes the 200 bucket (written_records=N) and opens a 404 bucket
  (failed_records=X), same shape.
- `replicated_data_size` and `duration` were not tracked before; threaded byte size through
  the sender result and added per-pass timing in the worker loop.
- Emission failures are swallowed-and-logged, never propagated into the replication path.

### Related issues

Closes #1374

### Points to confirm

- **Duration attribution on mixed-status passes:** a pass's measured duration is attributed
  once, to the lowest status code seen that pass (others get 0.0), to avoid double-counting.
  Exact for single-status passes and the 200→404 example. Is this the semantics you want, or
  would you prefer attribution by record share?
- **`status` cast:** kept as `u16` to match lifecycle (`lifecycle_task.rs`); note
  `forward_system_logger.rs` uses `i16` — pre-existing inconsistency, flagged not fixed here.
- **`replicated_data_size`** = bytes of successfully written records (0 for failure buckets).
- **`event_type`** = `"replication_sync"` (matches the `lifecycle_run` style) — rename if you'd prefer.
- Clippy 1.91 surfaces pre-existing repo-wide lints; CI doesn't gate on clippy, so left untouched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Periodically write aggregated replication diagnostics as system events to the `$system` bucket under `replications/<instance>/<replication-name>`, aggregated per status code and flushed on status change, idle window, or time cap, [PR-NNNN](https://github.com/reductstore/reductstore/pull/NNNN)
+- Periodically write aggregated replication diagnostics as system events to the `$system` bucket under `replications/<instance>/<replication-name>`, aggregated per status code and flushed on status change, idle window, or time cap, [PR-1417](https://github.com/reductstore/reductstore/pull/1417) by @DibbayajyotiRoy
 - Accept `x-reduct-content-length` as an alternative to `Content-Length` for browser streaming via the Fetch API, [PR-1411](https://github.com/reductstore/reductstore/pull/1411)
 - Pass attachment key-value maps to extensions instead of selecting a single attachment key, [PR-1383](https://github.com/reductstore/reductstore/pull/1383)
 - Add HTTP CRUD endpoints for lifecycle policies, [PR-1382](https://github.com/reductstore/reductstore/pull/1382)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Periodically write aggregated replication diagnostics as system events to the `$system` bucket under `replications/<instance>/<replication-name>`, aggregated per status code and flushed on status change, idle window, or time cap, [PR-NNNN](https://github.com/reductstore/reductstore/pull/NNNN)
 - Accept `x-reduct-content-length` as an alternative to `Content-Length` for browser streaming via the Fetch API, [PR-1411](https://github.com/reductstore/reductstore/pull/1411)
 - Pass attachment key-value maps to extensions instead of selecting a single attachment key, [PR-1383](https://github.com/reductstore/reductstore/pull/1383)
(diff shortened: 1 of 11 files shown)


DibbayajyotiRoy:
Hi @atimin  ,this implements #1374 per the spec in the issue thread. Ready for review.

What it does

Writes aggregated replication diagnostics to $system under replications/<instance>/<replication-name>, one event per status code, flushed on status change / 5s idle / 60s cap (reuses the ApiAuditEventAggregator timing).
Unified payload for success and failure: status, pending_records, written_records, failed_records, replicated_data_size, duration.
Mirrors the lifecycle SystemEventSink wiring. cargo test green, rustfmt clean.
Could you approve the workflow run so CI can verify?

A couple of design points I'd like your call on (also in the PR description):

On mixed-status passes, duration is attributed once to the lowest status code (others 0.0) to avoid double-counting exact for single-status passes and the 200 -> 404 example. OK, or prefer attribution by record share?
event_type = "replication_sync" (matching lifecycle_run) happy to rename.

DibbayajyotiRoy:
Hi @atimin 

Pushed a follow-up commit (bb36f2c): adds test coverage for the diagnostics worker-loop wiring (closes the codecov/patch gap) and de-flakes the Windows $system persistence test with deterministic polling instead of a fixed sleep. Could you re-approve the workflow run when you get a chance? Thanks!

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Periodically write aggregated replication diagnostics as system events to the `$system` bucket under `replications/<instance>/<replication-name>`, aggregated per status code and flushed on status change, idle window, or time cap, [PR-NNNN](https://github.com/reductstore/reductstore/pull/NNNN)
```
atimin: don't forget ID of task and "by @DibbayajyotiRoy"

atimin (changes requested):
Hi @DibbayajyotiRoy , it's  perfect!

However, I think I misled you by giving API Audit Aggregator as an example without mentioning that it forwards audit events from read-only replicas. 

https://www.reduct.store/docs/guides/disaster-recovery#read-only-replicas

Replications only work on the primary node, though. So, the process can be much simpler, and we can use the lifecycle policy module as an example. 

Is it ok for you to make another iteration?

DibbayajyotiRoy:
@atimin absolutely will try my best.

DibbayajyotiRoy:
@atimin done!!

and I think the CI is breaking because of environmental issues?

docker and http ones?

atimin (approved):

atimin:
@DibbayajyotiRoy Merged!

Issue #1389, opened 2026-05-22
Replication sender fails with TooEarly in cloud setup when storage backend is slow


mother-6000:
In a cloud setup with slower disk/S3 communication, replication sometimes fails with a `TooEarly` error because the record is still being written when the replication sender tries to read it.

## Error

```text
2026-05-22 12:47:57.169 (27200) [ERROR] -- reductstore/replication/replication_sender.rs:111 Failed to read record dev-1/Pablo05/sensor/gps/fix/1779453683528799: ReductError { status: TooEarly, message: "Record with timestamp 1779453683528799 in dev-1/Pablo05/sensor/gps/fix is still being written" }
```

## Context

- Environment: cloud deployment
- Storage/backend communication is slower there (disk/S3)
- This appears to create a race where replication reads a record before the write is fully completed

## Expected behavior

Replication should retry or otherwise tolerate records that are not yet fully written, instead of failing on `TooEarly`.

## Actual behavior

Replication sender logs an error and fails to read the record.

skaunov:
@atimin 

I got here via your call for contributors. I'd create draft PR to track the progress if there is no objections for trying to tackle this issue. 

While we're here, when there will be some development on this: how do I try it? A test instance, CI, or just to ask somebody to try and post the results?

atimin:
Hi @skaunov, thank you for wanting to help.  I'm afraid this is issue is pretty hard to reproduce since it is happening on a cloud setup with some load.  

However, I'm pretty sure we can imitate it in unit tests. It should be enough.

Should I assign it to you?  The task isn't urgent, but it would be great to deliver it in version 1.20 within two weeks.

skaunov:
I don't think it worth to bother assigning: related draft PR will do it for me.

PR #1415, opened 2026-06-05
feat(replication): Optimize replication task

FirasCh3:
Closes #

### Please check if the PR fulfills these requirements

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] CHANGELOG.md has been updated (for bug fixes / features / docs)

### What kind of change does this PR introduce?

(Bug fix, feature, docs update, ...)

### What was changed?

(Describe the changes)

### Related issues

(Add links to related issues)

### Does this PR introduce a breaking change?

(What changes might users need to make in their application due to this PR?)

PR #1419, opened 2026-06-06
fix: develop `ReplicationSender`

skaunov:

- [ ] Tests for the changes have been added (for bug fixes / features)

- Performance upgrade.
- development on `TooEarly` timeout

### What was changed?

Move `futures` into `dev-dependencies`.

Develop `ReplicationSender::run` to be fully concurrent on reading.

Internalize `async fn read_record` replacing it with the crate.

Move `CoreExtCfgParser` from function arguments to generic.

Also some lateral changes from active reading while I was getting to know around.

### Related issues

#1389 

No

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3022,6 +3022,7 @@ dependencies = [
  "crc64fast",
  "crossbeam-channel",
  "dlopen2",
+ "either",
  "flate2",
  "futures",
  "futures-util",
@@ -3049,6 +3050,7 @@ dependencies = [
  "tempfile",
  "test-log",
  "tokio",
+ "tokio-retry",
  "tower",
  "tower-http",
  "url",
@@ -4132,6 +4134,16 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn 2.0.117",
+]
+
+[[package]]
+name = "tokio-retry"
+version = "0.3.1"
+source = "git+https://github.com/skaunov/tokio-retry#9a37100ec56a689351a8d5d2c9be13523964e7fc"
+dependencies = [
+ "pin-project-lite",
+ "rand 0.10.1",
+ "tokio",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3022,7 +3022,6 @@ dependencies = [
  "crc64fast",
  "crossbeam-channel",
  "dlopen2",
- "either",
  "flate2",
  "futures",
  "futures-util",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4138,8 +4138,9 @@ dependencies = [
 
 [[package]]
 name = "tokio-retry"
-version = "0.3.1"
-source = "git+https://github.com/skaunov/tokio-retry#9a37100ec56a689351a8d5d2c9be13523964e7fc"
+version = "0.3.2"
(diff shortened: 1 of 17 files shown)


atimin (changes requested):
Hi @skaunov , thank you for your PR.

Unfortunately, it won't work..  the current design doesn't allow to parallelize the communication with the target bucket. We're working on it here:https://github.com/reductstore/reductstore/pull/1415

However even, if we could optimize this part, it wouldn't solve the problem anyway. The reason is that the issue happens on the storage engine level it just needs more time to sync data between file system and remote storage.

My proposal is to elaborate the `ReplicationSender::read_record`.  We can increase the timeout for the remote backend, and make it incremental:

1) filesystem backend: 5ms -> 10ms -> 20ms
2) remote storage: 50ms -> 100ms -> 200ms

What do you think?

skaunov:
So am I correct that `attempts` are not needed at all as `record` eventually gets to a final state?

Basically I started from moving/stepping in that direction. I had a question about that actually: how this part can know about the kind of remote back-end (to distinguish FS from S3)? 

Tell me, pls, why is there *so much* of `dyn`? I know two usual reasons: interfacing with something and extensibility. I did not stumble upon interfacing part (maybe I just went other place). Maybe you have a link on this choice?

atimin:
> So am I correct that attempts are not needed at all as record eventually gets to a final state?

They are necessary, because there are no guaranties that record can get the final state. A network issue,  the client crashed etc. 

> Basically I started from moving/stepping in that direction. I had a question about that actually: how this part can know about the kind of remote back-end (to distinguish FS from S3)?

Currently you can't, we need to change the ReplicationSender::new function. it should receive the Cfg struct. It's global configuration, there you can find the backend_config field. 

> Tell me, pls, why is there so much of dyn? I know two usual reasons: interfacing with something and extensibility. I did not stumble upon interfacing part (maybe I just went other place). Maybe you have a link on this choice?

'So much' is a very subjective judgment. We use  traits, when we need to enable\disable a component,  or it must change behavior depending on the instance role e.g. primary and read-only replica have very different functionalities. It also a way to mock objects in tests. Usually there is a reason to do so.

atimin:
Hi @skaunov,
Thanks for your efforts, but I failed to understand what your PR is doing.  Why did you rewrite the run function? 

And also why did you touch not related part of code? It takes time to review but doesn't add value.

skaunov:
It turned out to be such a good object to both wrap my head around some parts of the system & to do few experiments with Tokio. Do you see any defects in mine development?

I never worked with Reduct Store before, so had to move around --- could not help to check some things. My part is just to offer: I feel like ditching that argument which actually been only a generic is a good (as in adding value) thing. So it's just an opinion expressed in code. :shrug:

atimin:
No problem. I don't mind discussing new things.  Could you explain the idea and why it's better than the previous implementation?   At least the previous variant looked easier to understand and maintain.

skaunov:
The idea here is that quicker batches could be already written while `TooEarly` transactions are still awaiting to come through. So more efficiency, very much more depending on how unlucky they line up in current version. Plus it does it without spawning a task (as it seems no need for that as I feel it just/mostly juggles requests underneath).

I've added few comments to help follow along: does it helps? Am glad to answer questions on this or to expand some parts.

PR #1431, opened 2026-06-10
Feat/usage statistics

### LINKED ISSUE #1428: Add usage statistics to /usage/<node-name> as flat JSON records
## Description
We need a unified way to expose usage statistics for an instance/account in the system stats bucket.

Part of this information is already available from `limits`, while the rest can be derived from the storage itself. Instead of forcing users to collect these values from different places, ReductStore should write usage statistics as flat JSON records into dedicated `$system/usage/<node-name>` entries, following the same general approach as audit and other system stats.

Each node should write its own usage records:
- primary node → `$system/usage/<primary-node>`
- read-only replica → `$system/usage/<replica-node>`

Read-only replicas must forward their usage records upstream the same way audit/system stats are forwarded today.

## Requested metrics
- number of API calls
- ingest/write traffic
- read/egress/download traffic
- number of records written
- number of records read
- total data stored
- number of buckets
- number of entries
- number of blocks

## Format
- bucket/entry: `$system/usage/<node-name>`
- content type: JSON
- timestamp: stored in the standard record header/query parameter (`ts`)
- payload: flat JSON for simpler processing

## JSON example
```json
{
  "duration": 60.0,
  "api_calls": 1250,
  "write_bytes": 1844674407,
  "read_bytes": 9234172031,
  "records_written": 15834022,
  "records_read": 93450211,
  "storage_bytes": 43872123904,
  "bucket_count": 47,
  "entry_count": 1823,
  "block_count": 128442
}
```

## Metric semantics
- `duration` — length of the measurement interval in seconds
- `api_calls` — number of API calls during the interval
- `write_bytes` — bytes ingested during the interval
- `read_bytes` — bytes read/downloaded during the interval
- `records_written` — number of records written during the interval
- `records_read` — number of records read during the interval
- `storage_bytes` — snapshot of stored data size at `ts`
- `bucket_count` — snapshot of bucket count at `ts`
- `entry_count` — snapshot of entry count at `ts`
- `block_count` — snapshot of block count at `ts`

## Replica forwarding
Read-only replicas must generate the same usage records and forward them to the primary, using separate node-specific entries.

For example:
- `$system/usage/primary-node`
- `$system/usage/replica-a`
- `$system/usage/replica-b`

This should follow the same forwarding model as audit events and other system statistics.

## Implementation note
- some metrics can be sourced from existing `limits` accounting
- the rest can be derived from the storage layer
- traffic and record counters should be interval-based
- storage/bucket/entry/block values should be point-in-time snapshots

## Why this is needed
- monitor real usage
- estimate capacity and costs
- track growth over time
- include usage served by replicas
- simplify observability and billing/reporting

## Acceptance criteria
- usage statistics are written to `$system/usage/<node-name>`
- payload is flat JSON
- timestamp uses the standard `ts` record header/query parameter
- stats include:
  - `duration`
  - `api_calls`
  - `write_bytes`
  - `read_bytes`
  - `records_written`
  - `records_read`
  - `storage_bytes`
  - `bucket_count`
  - `entry_count`
  - `block_count`
- primary and replicas use the same schema
- read-only replicas forward their usage records upstream
- behavior and field semantics are documented clearly

DibbayajyotiRoy:
Closes #1428

Feature.

### What was changed?
Adds usage statistics reporting. Each node now writes its own usage metrics as flat-JSON records into `$system/usage/{instance}/total` every 60 seconds, reusing the existing system-event mechanism (the same pathway as lifecycle and replication diagnostics). Read-only replicas forward their usage records upstream the same way other system events are forwarded, so the primary gets a consolidated, per-node view.

The payload has 9 flat fields:
- **Interval counters** (reset each interval): `write_bytes`, `read_bytes`, `records_written`, `records_read`
- **Point-in-time snapshots**: `storage_bytes`, `bucket_count`, `entry_count`, `block_count`
- **`duration`**: measured elapsed seconds of the interval

Implementation notes:
- Counters are tallied at the storage-engine choke points (`StorageEngine::begin_write` for writes, `RecordReader::try_new` for reads), so all traffic is counted uniformly regardless of source (HTTP / replication / zenoh), per the issue's intent. They live as `Arc<UsageCounters>` on `Cfg` (alongside `limits_config`) since `Cfg` already flows to both points; drained via atomic `swap(0)` so increments during a flush roll into the next interval with no loss.
- The four snapshots come from a single bucket walk per tick (reusing `bucket.info()` / `FullBucketInfo`), the same cost as one `GET /api/v1/info`.
- `api_calls` is intentionally **not** included — API calls are already covered by audit logs.
- The 60s interval is hard-coded (no new config option); gating reuses the existing `RS_SYSTEM_EVENTS_ENABLED` flag.
- Final usage event is flushed on shutdown before storage sync.

### Design notes to confirm
A few counting-boundary choices I'd like to confirm are intended:
1. **Self-counting:** system-event writes (audit / lifecycle / usage itself) count as write traffic in the next interval, consistent with "all traffic counts uniformly." So `write_bytes` won't read zero on an otherwise-idle instance.
2. **Count-at-creation:** bytes are tallied when the reader/writer is created (full `content_size`), so an aborted mid-stream transfer still counts in full. Metadata-only query results count neither records nor bytes.
3. **Snapshot scope:** `bucket_count` / `entry_count` / `block_count` use `bucket.info()`, which (like the public API) excludes hidden `$meta` entries from entry/block counts while `storage_bytes` includes their size; `$system` itself is included in the walk, matching `GET /api/v1/info`.

No. New records appear under `$system/usage/{instance}/total` when system events are enabled; no existing behavior changes.

### Other information:
- Entry path is `$system/usage/{instance}/total`; the `total` entry name leaves room for per-bucket usage entries (e.g. `{instance}/bucket-1`) as a future extension.
- 9 new tests cover counter drain/no-loss, snapshot sums, the 9-field event, emission-failure resilience, shutdown flush, and the entry-path end-to-end.
- `cargo build` / `cargo test` clean: 2056 passed, 0 failed (includes the merge with current `main`).
- Happy to follow up with a docs PR documenting the `$system/usage/{instance}/total` entry and its 9 fields if wanted.


DibbayajyotiRoy:
The Build Binaries (aarch64-apple-darwin) check is failing in setup-protoc, not in the code — fork PRs don't receive repo secrets, so the protoc installer falls back to unauthenticated GitHub API calls and hits the shared-runner rate limit (it fails before anything compiles). I tried re-triggering with an empty commit but it landed on a throttled IP again. Could you re-run the failed jobs when you have a moment? Builds and the full test suite pass locally (2056 passed, 0 failed). Happy to do anything else that helps.

atimin:
@DibbayajyotiRoy , unfortunately, the CI/CD uses secrets that are linked to our internal infrastructure. That's why it doesn't work well for external forks. In my opinion, the green "Generate Code Coverage" job would be good criteria that the code is ready for review.  However, it must also be based on the best effort approach because sometimes it isn't possible to cover all scenarios.

DibbayajyotiRoy:
Thanks for the context — understood about the fork secrets. Agreed on the coverage job as the signal: in the last run it executed the full suite and failed only on lock_file::tests::test_secondary_instance_waits, a pre-existing timing test that flaked under llvm-cov instrumentation (unrelated to this diff; passes locally). I've re-triggered the workflow to get it green. The new usage module itself has 9 dedicated tests covering the counters, snapshot sums, the event payload, emission-failure resilience, shutdown flush, and the entry path end-to-end.
@atimin

reductstore/src/api/components.rs
```diff
@@ -40,6 +41,8 @@ pub struct Components {
     pub(crate) query_link_cache: AsyncRwLock<Cache<String, Arc<Mutex<BoxedReadRecord>>>>,
     pub(crate) audit_logger: Arc<AsyncRwLock<Box<dyn LogSystemEvent + Send + Sync>>>,
     pub(crate) limits: BoxedLimits,
+    /// Periodic usage statistics task; `None` when system events are disabled.
+    pub(crate) usage_stats_task: AsyncRwLock<Option<UsageStatsTask>>,
```
atimin: I would prefere if we follow our previous approach we used for audit_logger.

1. We usage_stat_logger in components
2. The logger has an aggregator with the task.

reductstore/src/cfg.rs
```diff
@@ -319,6 +325,7 @@ impl<EnvGetter: GetEnv, ExtCfg: ExtCfgBounds> CfgParser<EnvGetter, ExtCfg> {
             rw_lock_config: Self::parse_rw_lock_config(&mut env),
             engine_config: Self::parse_storage_engine_config(&mut env),
             limits_config: Self::parse_limits_config(&mut env),
+            usage_counters: Arc::new(UsageCounters::default()),
```
atimin: The structure is purely to store and forward the configuration through other components.
Usage counters are a component.

reductstore/src/lib.rs
```diff
@@ -13,3 +13,4 @@ pub mod lock_file;
 pub mod replication;
 pub mod storage;
 pub mod syslog;
+pub mod usage;
```
atimin: the usage module must be in storage

atimin (changes requested):
You didn't request a review, but I noticed the approach is different from what we used before, so let's discuss it first.

reductstore/src/api/components.rs
DibbayajyotiRoy: Agreed, I'll mirror the audit_logger pattern: Components gets usage_stat_logger: Arc<AsyncRwLock<BoxedSystemLogger>>, and the logger owns an aggregator that spawns the 60s worker internally (like ApiAuditEventAggregator). The worker will flush the partial interval and exit when the logger is dropped, so the explicit StateKeeper::stop_usage_stats_task() and the launcher shutdown hook go away too.

reductstore/src/cfg.rs
DibbayajyotiRoy: Agreed, I'll take the counters out of Cfg. One question on where the component should live: the read choke point (RecordReader::try_new) only has access to the BlockManager, so the options are (1) a module-level singleton in storage::usage, same pattern as FILE_CACHE, or (2) threading Arc<UsageCounters> from the engine down through Bucket/Entry/BlockManager constructors like InFlightIoLimiter. (1) is a much smaller diff; (2) keeps it injectable. Which do you prefer?

reductstore/src/lib.rs
DibbayajyotiRoy: Will do moving it to storage::usage.

DibbayajyotiRoy:
@atimin I deeply apologise to not ask for review. I am assuring you, I won't repeat the same mistake again.

reductstore/src/cfg.rs
atimin: FILE_CACHE is a source of many problems since it's a singleton. I would prefer to use Arc or even Arc<RwLock> if necessary. Consider to use  the use_stat_logger component instead of the usage_counters. Very important we don't lock it for a long time, but we already mitigate it having the tokio task + channel.

DibbayajyotiRoy:
moved the module under storage, and reworked the periodic task into a usage_stat_logger + aggregator that owns the worker and stops on drop (channel-close), mirroring audit_logger dropped the explicit stop plumbing. Holding the counter relocation on your answer above.
@atimin

DibbayajyotiRoy:
@atimin All three points addressed in the latest commit:

Counters off Cfg — UsageCounters is now a component owned by the StorageEngine (created in its builder) and passed as Arc down through the Bucket/Entry/BlockManager constructors, the same way Cfg and the IO limiter already flow. No singleton. The hot path stays lock-free (relaxed atomic increments); only the aggregator's own task takes the logger lock, once per interval.
audit_logger pattern — Components now holds usage_stat_logger, built by build_usage_logger() mirroring build_audit_logger(). The logger owns a UsageEventAggregator whose worker flushes every 60s and shuts down via channel-close when the logger is dropped, so the explicit stop plumbing in StateKeeper/launcher is gone.
Module location — moved to storage::usage.
No behavior change: same event schema, entry path, and 60s interval. Full suite green (2056 passed). Ready for another look when you have time.

reductstore/src/api/components.rs
```diff
@@ -40,6 +40,13 @@ pub struct Components {
     pub(crate) query_link_cache: AsyncRwLock<Cache<String, Arc<Mutex<BoxedReadRecord>>>>,
     pub(crate) audit_logger: Arc<AsyncRwLock<Box<dyn LogSystemEvent + Send + Sync>>>,
     pub(crate) limits: BoxedLimits,
+    /// Usage statistics logger; owns the periodic flush task and stops it on
+    /// drop (a disabled logger when system events are off). Unlike
+    /// `audit_logger`, usage events are produced by the task's own 60s timer
+    /// rather than logged through this handle, so the field is held only to
+    /// keep the task alive for the server's lifetime and is never read.
+    #[allow(dead_code)]
+    pub(crate) usage_stat_logger: Arc<AsyncRwLock<Box<dyn LogSystemEvent + Send + Sync>>>,
```
atimin: The logger isn't being used, and the block manager is using old counters. I suppose it's not ready yet, ritgh?

DibbayajyotiRoy: Good catches — both fixed.

Counters: there was no functional disconnect (the aggregator drained the engine's counters and the e2e test confirms non-zero reads/writes land in the event), but it read as disconnected because the engine minted the Arc<UsageCounters> and the logger pulled it back via an accessor. Now there's one Arc<UsageCounters> created in build() and passed explicitly to both the storage engine (with_usage_counters, threaded to the block managers) and the usage aggregator that drains it.

Logger not used: you're right — forcing usage into the audit_logger shape left a no-op log_event and a dead field, because nothing logs to a timer-driven task the way middleware logs to audit. I've made it a concrete UsageEventAggregator (dropped the dyn LogSystemEvent), stored as Option<UsageEventAggregator> with no #[allow(dead_code)], and it's stopped on shutdown via StateKeeper (which also makes the final flush deterministic instead of relying on drop). Full suite green.

DibbayajyotiRoy:
@atimin 

The coverage job was cancelled at the 30-min timeout, not a test failure — the log shows the S3 build cache failed to authenticate (Could not load credentials from any providers), so it did a full cold instrumented compile (26m43s) and blew the budget. The previous coverage run did it in ~3 min with a warm cache. Verified locally: the full suite under --features fs-backend,ci runs in 24s, 2057 passed / 0 failed, no hang. Could you re-run the job?

atimin (approved):
Perfect! Thank you!

PR #1474, opened 2026-06-22
Feat/per bucket usage stats

### LINKED ISSUE #1468: Usage stats: add per-bucket usage plus entry read/write and record counters
## Problem statement

Current usage statistics do not provide enough detail for bucket-level analysis and workload characterization.

We need more visibility not only into total usage, but also into:
- per-bucket usage
- how many entries were written during the stats interval
- how many entries were read during the stats interval
- total and per-bucket record counts

Without these metrics, it is harder to understand dataset growth, access patterns, and bucket-level load.

## Proposed solution

Improve usage statistics so they include:

- per-bucket usage alongside total usage
- `written_entries` — number of entries written during the stats interval
- `read_entries` — number of entries read during the stats interval
- `record_count` — total record count and per-bucket record count

## Scope

The implementation should:
- extend the usage stats output format
- keep total usage reporting
- add bucket-level usage and record count metrics
- add interval counters for written and read entries
- document the new fields

## Expected outcome

Users can inspect storage growth and access behavior at both total and bucket levels, and can better understand how many entries and records are being written and read over time.

DibbayajyotiRoy:
Closes #1468.

## Summary

Extends the usage-stats feature (#1431) with per-bucket granularity and two new metrics. Each user bucket now gets its own event at `$system/usage/{instance}/{bucket}` alongside the existing `total`, using the same flat schema.

## New fields

Added to both the `total` event and each per-bucket event:

| Field | Type | Meaning |
|-------|------|---------|
| `written_entries` | interval | Distinct entries written to during the interval |
| `read_entries` | interval | Distinct entries read from during the interval |
| `record_count` | snapshot | Total record count (and per-bucket record count) |

## Notes

- **Total reporting is unchanged.** Per-bucket counters are additive alongside the existing global atomics, so the `total` event behaves exactly as before.
- **`written_entries` / `read_entries` use composite keying.** An entry is identified by bucket + name, so the same entry name in two different buckets counts as **2** in the instance total (per-bucket counts are always exact). This matches "how many entries are used for writing and reading data." It's a one-line change to name-only keying if you'd prefer that instead.
- **System buckets are excluded** from per-bucket events. `$`-prefixed names (`$system`, `$audit`) are invalid as entry-path segments, and telemetry shouldn't report on itself.

## Tests

| Test | Verifies |
|------|----------|
| `distinct_entries_count_unique_names_not_operations` | 5 writes to one entry → `written_entries=1`, `records_written=5` |
| `distinct_entries_are_keyed_by_bucket_and_name` | Same name in two buckets → total = 2 |
| `per_bucket_counts_isolate_by_bucket` | Counts attributed to the correct bucket |
| `drained_buckets_are_removed_from_the_map` | Deleted buckets don't grow the map unbounded |
| `flush_emits_total_and_per_bucket_events` | One event per bucket plus `total` |
| `flush_excludes_system_buckets_from_per_bucket_events` | `$`-prefixed buckets skipped |

16 usage tests pass; `cargo fmt --all -- --check` clean.

### DIFF
--- a/reductstore/src/storage/entry/io/record_reader.rs
+++ b/reductstore/src/storage/entry/io/record_reader.rs
@@ -81,7 +81,8 @@ impl RecordReader {
         };
 
         bm.usage_counters()
-            .count_read(bm.bucket_name(), bm.entry_name(), content_size);
+            .count_read(bm.bucket_name(), bm.entry_name(), content_size)
+            .await?;
 
         Ok(Self {
             meta,
--- a/reductstore/src/storage/entry/io/record_reader.rs
+++ b/reductstore/src/storage/entry/io/record_reader.rs
@@ -80,7 +80,8 @@ impl RecordReader {
             None
         };
 
-        bm.usage_counters().count_read(content_size);
+        bm.usage_counters()
+            .count_read(bm.bucket_name(), bm.entry_name(), content_size);
 
         Ok(Self {
             meta,
--- a/reductstore/src/storage/usage.rs
+++ b/reductstore/src/storage/usage.rs
@@ -15,49 +15,115 @@ mod usage_event_payload;
 
 pub(crate) use usage_aggregator::UsageEventAggregator;
 
+use std::collections::{HashMap, HashSet};
 use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::Mutex;
 
 /// Interval traffic counters incremented at the storage engine choke points.
 ///
-/// Plain relaxed atomics: the counters are independent tallies, no ordering
-/// between them is required. `drain` swaps each counter to zero, so
-/// increments that race with a flush roll into the next interval instead of
-/// being lost.
+/// The instance-wide tallies are plain relaxed atomics: independent counters,
+/// no ordering between them is required. Per-bucket traffic — including each
+/// bucket's set of distinct entries written to / read from — is kept in
+/// `per_bucket` behind a single mutex so each `count_*` updates it in the same
+/// critical section. `drain` resets everything, so increments that race with a
+/// flush roll into the next interval instead of being lost.
 #[derive(Debug, Default)]
 pub(crate) struct UsageCounters {
     write_bytes: AtomicU64,
     read_bytes: AtomicU64,
     records_written: AtomicU64,
     records_read: AtomicU64,
+    per_bucket: Mutex<HashMap<String, BucketTraffic>>,
 }
 
-/// Counter values drained for one flush interval.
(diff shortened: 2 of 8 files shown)


DibbayajyotiRoy:
Heya! @atimin please review the PR whenever you have time, and please tell your changes, which I will do if needed, I currently cannot find the usual review option to request review , that's why commenting it!

reductstore/src/storage/usage.rs
```diff
@@ -15,49 +15,115 @@ mod usage_event_payload;
...
+/// `per_bucket` behind a single mutex so each `count_*` updates it in the same
+/// critical section. `drain` resets everything, so increments that race with a
+/// flush roll into the next interval instead of being lost.
 #[derive(Debug, Default)]
 pub(crate) struct UsageCounters {
     write_bytes: AtomicU64,
     read_bytes: AtomicU64,
     records_written: AtomicU64,
     records_read: AtomicU64,
+    per_bucket: Mutex<HashMap<String, BucketTraffic>>,
```
atimin: please, use use crate::core::sync::AsyncRwLock , since it's used in a asynchronous context. 
of function count_write\count_read must be asyncronouse.

reductstore/src/storage/entry/io/record_reader.rs
```diff
@@ -80,7 +80,8 @@ impl RecordReader {
             None
         };
 
-        bm.usage_counters().count_read(content_size);
+        bm.usage_counters()
```
atimin, line 83: I think we could do the same in the RecordWriter ,  currently we're doing it in StorageEngine.begin_write and it's a bit confusing.

atimin (changes requested):
Hi @DibbayajyotiRoy , thank you very much for the contriubtion. It works perfectly, just few remarks about implementation.

DibbayajyotiRoy:
Both addressed: per_bucket is now AsyncRwLock with async count_write/count_read/drain, and write counting moved into RecordWriter::try_new, symmetric with the read counting in RecordReader. To avoid holding a block lock across the count .await, I extract an owned Arc<UsageCounters> while the block-manager guard is held and count after the block locks release.
One choice I'd like your call on: count_write/count_read currently propagate lock errors with ?, which means they're now on the write/read IO path. Since usage counting is best-effort telemetry, I'd lean toward swallowing-and-logging lock errors inside count_* instead, so a telemetry failure can never surface as a failed write or read (matching how the aggregator already swallows emission errors). Want me to switch to swallow-and-log, or keep propagation for consistency?
@atimin

atimin (approved):
Perfect! @DibbayajyotiRoy , thank you for the contribution!

PR #1481, opened 2026-06-26
feat(syslog): capture log messages to $system/logs/<instance> (#1467)

### LINKED ISSUE #1467: System events: record log messages to $system/logs/<instance> with configurable log level
## Problem statement

ReductStore currently does not persist internal log messages into the system event stream.

This makes it harder to inspect instance behavior through the same data access path used for other operational signals, and it limits observability for debugging distributed deployments.

## Proposed solution

Record log messages into the `$system/logs/<instance>` entry.

Add an environment setting:

- `RS_SYSTEM_EVENTS_LOG_LEVEL`

This setting defines the minimum log level that should be written into the system log entry.

Default:
- `WARN`

## Scope

The implementation should:
- write internal log messages to `$system/logs/<instance>`
- filter stored messages by `RS_SYSTEM_EVENTS_LOG_LEVEL`
- keep the default behavior at `WARN`
- document the new setting and system entry behavior

## Expected outcome

Users can query operational logs directly from `$system/logs/<instance>` and control verbosity with `RS_SYSTEM_EVENTS_LOG_LEVEL`, making system diagnostics easier in single-node and fleet setups.

DibbayajyotiRoy:
Closes #1467.

## Summary
Captures the system's own log messages into `$system/logs/<instance>/messages`, queryable like other `$system` events. Opt-in via `RS_SYSTEM_EVENTS_LOG_LEVEL` (off by default).

## Approach (per our discussion)
- **`reduct-base` stays abstract.** The logger gains a generic sink hook (`LogSnapshot` + `set_log_sink`) with zero knowledge of reductstore types. All `$system` mapping lives in reductstore and registers into the hook, so `reduct-base` keeps no dependency on `reductstore`.
- **Console output is unconditional; only capture is guarded.** A log emitted during a capture write still prints to stdout (a `tokio::task_local!` reentrancy guard suppresses only re-enqueue, not the console `println!`), so logging errors during logging stay visible.
- Node-local only (non-replica), bounded channel with drop-on-overflow so capture never blocks the server.

## For your call
- `status` carries the level code (1=ERROR … 5=TRACE) and level label, reusing the shared flat-schema field. Happy to use a separate field if you'd rather not overload `status`.
- **max_level limitation:** `set_max_level` is still derived from the console level, so a persist level *more verbose* than the console level won't capture the extra records. Left out per our "start simple" plan; the fix is `set_max_level(max(console, persist))` at startup if you want it.

## Tests
- `consumer_does_not_loop_when_write_emits_logs` — reentrancy: a write that emits logs writes exactly its seeds, no amplification, terminates.
- `console_prints_even_when_sink_reenters` — console visibility preserved during capture.
- Plus level filtering, overflow drop+count, round-trip to the entry path, gating, schema mapping.

### DIFF
--- a/reduct_base/src/logger.rs
+++ b/reduct_base/src/logger.rs
@@ -8,9 +8,21 @@ use std::sync::{Arc, LazyLock, RwLock};
 use std::time::{SystemTime, UNIX_EPOCH};
 use thread_id;
 
-static LOGGER: Logger = Logger;
+static LOGGER: Logger = Logger {
+    sink: RwLock::new(None),
+};
 
-pub struct Logger;
+/// A registered log sink: receives an owned [`LogSnapshot`] for every record
+/// that reaches the logger. It must be non-blocking and must not panic.
+pub type LogSink = Arc<dyn Fn(LogSnapshot) + Send + Sync>;
+
+pub struct Logger {
+    /// Optional capture sink, installed via [`Logger::init_with_sink`]. Held as
+    /// interior-mutable state on the single static logger rather than as a
+    /// separate global.
+    sink: RwLock<Option<LogSink>>,
+}
+
 static PATHS: LazyLock<RwLock<BTreeMap<String, Level>>> =
     LazyLock::new(|| RwLock::new(BTreeMap::new()));
 
@@ -32,8 +44,8 @@ pub struct LogSnapshot {
     pub message: String,
 }
 
-impl LogSnapshot {
-    fn from_record(record: &Record) -> Self {
+impl From<&Record<'_>> for LogSnapshot {
+    fn from(record: &Record) -> Self {
         let timestamp = SystemTime::now()
             .duration_since(UNIX_EPOCH)
             .map(|d| d.as_micros() as u64)
@@ -47,23 +59,6 @@ impl LogSnapshot {
             message: format!("{}", record.args()),
         }
     }
-}
-
-/// A registered log sink: receives an owned [`LogSnapshot`] for every record
-/// that reaches the logger. It must be non-blocking and must not panic.
-pub type LogSink = Arc<dyn Fn(LogSnapshot) + Send + Sync>;
-
-static LOG_SINK: RwLock<Option<LogSink>> = RwLock::new(None);
-
-/// Register (or replace) the global log sink. Called once at startup by the
-/// component that persists logs; `reduct_base` itself never inspects the sink.
-pub fn set_log_sink(sink: LogSink) {
-    *LOG_SINK.write().unwrap() = Some(sink);
-}
-
-/// Remove the global log sink (used on shutdown and in tests).
-pub fn clear_log_sink() {
-    *LOG_SINK.write().unwrap() = None;
 }
 
 /// Parse a single log level name (case-insensitive) into a [`Level`].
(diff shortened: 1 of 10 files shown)


DibbayajyotiRoy:
@atimin Heya! please review the PR, i currently made, I will fix if anything needed

reductstore/src/cfg/system_events.rs
```diff
@@ -42,6 +48,9 @@ impl<EnvGetter: GetEnv, ExtCfg: ExtCfgBounds> CfgParser<EnvGetter, ExtCfg> {
                 env.get_optional::<String>("RS_SYSTEM_EVENTS_ENABLED"),
                 DEFAULT_SYSTEM_EVENTS_ENABLED || has_lifecycles,
             ),
+            log_level: env
+                .get_optional::<String>("RS_SYSTEM_EVENTS_LOG_LEVEL")
+                .and_then(|level| parse_log_level(&level)),
```
atimin: The log level should be set to 'WARN' by default.

reduct_base/src/logger.rs
```diff
@@ -4,14 +4,83 @@
...
+    /// Microseconds since the Unix epoch, captured when the record was logged.
+    pub timestamp: u64,
+    pub level: Level,
+    pub target: String,
+    pub file: Option<String>,
+    pub line: Option<u32>,
+    pub message: String,
+}
+
+impl LogSnapshot {
```
atimin: It's more idiomatic to use the 'From<>;' trait.

reduct_base/src/logger.rs
```diff
@@ -4,14 +4,83 @@
...
+
+/// A registered log sink: receives an owned [`LogSnapshot`] for every record
+/// that reaches the logger. It must be non-blocking and must not panic.
+pub type LogSink = Arc<dyn Fn(LogSnapshot) + Send + Sync>;
+
+static LOG_SINK: RwLock<Option<LogSink>> = RwLock::new(None);
+
+/// Register (or replace) the global log sink. Called once at startup by the
+/// component that persists logs; `reduct_base` itself never inspects the sink.
+pub fn set_log_sink(sink: LogSink) {
```
atimin: we don't need additional  accessors, let's use Logger::init_with_sink.

reduct_base/src/logger.rs
```diff
@@ -4,14 +4,83 @@
...
+            message: format!("{}", record.args()),
+        }
+    }
+}
+
+/// A registered log sink: receives an owned [`LogSnapshot`] for every record
+/// that reaches the logger. It must be non-blocking and must not panic.
+pub type LogSink = Arc<dyn Fn(LogSnapshot) + Send + Sync>;
+
+static LOG_SINK: RwLock<Option<LogSink>> = RwLock::new(None);
```
atimin: I believe it can be a field of Logger structure.

atimin (changes requested):
Hi @DibbayajyotiRoy , generally the solution is working well. I like it.  Check my comments, please.

I would also change the output:

Image

status must be 200 - HTTP OK
it's also helpful to label messages per severity for better filtering.

DibbayajyotiRoy:
All five addressed: status is now 200 with severity as a queryable level label; default log level is WARN (so capture is on by default, opt-out via RS_SYSTEM_EVENTS_LOG_LEVEL=OFF); LogSnapshot uses From<&Record>; sink installs via Logger::init_with_sink (no separate accessors); and the sink is now a field of Logger rather than a static.
One choice I'd like your call on: the level label is currently added in LocalSystemLogger by reading it from the event payload, which avoids touching all the SystemEvent construction sites. It works and is self-contained (logs are node-local so only the local writer matters), but if you'd rather have an explicit SystemEvent.labels field instead of inferring from the payload, I'm happy to switch. Let me know.

@atimin

atimin (changes requested):
@DibbayajyotiRoy , It's perfect, but CHANGELOG must be updated.

DibbayajyotiRoy:
@atimin done!

atimin (approved):

PR #1496, opened 2026-06-30
Feat/unify syslog

### LINKED ISSUE #1485: Refactor: unify $system event logging under syslog with one system_event_logger
## Summary
Refactor system event handling around a single `system_event_logger` component and move all related runtime logic into a dedicated `syslog` module.

Today `$system` stores several kinds of system event records (`audit`, `usage`, lifecycle results, replication events, error logs, etc.). In the current implementation these concerns are split across multiple modules, and each statistic/event producer has its own logger-ish component plus separate aggregator/task logic.

Current branch / latest state for reference:
- https://github.com/DibbayajyotiRoy/reductstore/tree/feat/system-logs

## Goal
Unify system-event writing and make `syslog` the single home for:
- event model;
- event sink / logger;
- local-vs-forwarded persistence;
- aggregators;
- counters / periodic flush workers;
- payload structs for all `$system` event types.

Configuration is **not** part of this refactor and should remain under `cfg`.

### 1. One `system_event_logger` component for all `$system` events
Instead of separate logger components per event class, introduce one shared component responsible only for accepting normalized system events and persisting/forwarding them.

Responsibilities:
- resolve the final `$system/<kind>/<instance>/<entry>` path;
- write locally on primary / standalone instances;
- forward to primary on readonly replicas;
- apply common labels / serialization / error handling;
- expose one API for all producers.

This component should not contain audit/usage/lifecycle-specific aggregation rules. It should be a generic sink.

### 2. Move all event production logic into `syslog`
All logic that decides **when** and **how** to emit system events should live in `syslog`, including:
- aggregators;
- counters;
- periodic flush workers;
- event-specific payload builders;
- routing / naming conventions for system entries.

Other modules (`api`, `lifecycle`, `replication`, `storage`) should only publish raw facts or call the `syslog` API, not own separate system-event pipelines.

## Proposed `syslog` module structure
Repository convention constraints:
- **do not use `mod.rs`**;
- **do not move config out of `cfg`**.

A structure along these lines would keep responsibilities clear:

```text
src/syslog.rs
src/syslog/event.rs                 # SystemEvent, common metadata, serialization
src/syslog/path.rs                  # build $system/<kind>/<instance>/<entry> paths
src/syslog/sink.rs                  # SystemEventSink / trait used by producers
src/syslog/system_event_logger.rs   # single logger facade
src/syslog/local_writer.rs          # local write implementation
src/syslog/forward_writer.rs        # replica -> primary forward implementation
src/syslog/registry.rs              # optional: build/start/stop all syslog tasks in one place

src/syslog/payload.rs
src/syslog/payload/
  audit.rs
  lifecycle.rs
  replication.rs
  usage.rs
  log.rs
  stats.rs

src/syslog/aggregate.rs
src/syslog/aggregate/
  audit.rs
  usage.rs
  replication.rs
  lifecycle.rs
  counters.rs

src/syslog/capture.rs
src/syslog/capture/
  logs.rs
```

## Suggested API shape
The rest of the codebase should depend on something small, for example:

```rust
trait SystemEventLogger {
    async fn emit(&self, kind: SystemEventKind, event: SystemEvent) -> Result<(), ReductError>;
}
```

Or, if routing is already encoded in the event:

```rust
trait SystemEventLogger {
    async fn emit(&self, event: SystemEvent) -> Result<(), ReductError>;
}
```

Then event-specific producers can live behind `syslog::aggregate::*` and emit normalized events through the same sink.

## Mapping from current branch to proposed structure
From `feat/system-logs`, the current pieces roughly map like this:

- `src/syslog.rs` -> slimmer top-level entry point + builder/wiring surface
- `src/syslog/local_system_logger.rs` -> `src/syslog/local_writer.rs`
- `src/syslog/forward_system_logger.rs` -> `src/syslog/forward_writer.rs`
- `src/syslog/log_capture.rs` -> `src/syslog/capture/logs.rs`
- `src/syslog/log_event_payload.rs` -> `src/syslog/payload/log.rs`
- `src/api/audit/aggregator.rs` -> `src/syslog/aggregate/audit.rs`
- `src/storage/usage/usage_aggregator.rs` -> `src/syslog/aggregate/usage.rs`
- `src/cfg/system_events.rs` -> **stays where it is**

## Why this refactor is worth it
- one canonical system-event pipeline instead of several slightly different ones;
- easier to add new `$system` event types;
- less duplication between audit / usage / logs / lifecycle / replication event handling;
- clearer ownership: producers in `syslog`, persistence in one logger;
- simpler tests around path generation, forwarding, aggregation, and payload encoding.

## Non-goals
- changing the external `$system` record format unless needed;
- changing event semantics during the first refactor;
- moving configuration out of `cfg`;
- mixing business logic back into unrelated modules.

## Acceptance criteria
- one shared system event logger is used for all `$system` writes;
- event-specific aggregation / counters / flush tasks live under `syslog`;
- audit and usage logic no longer own their own parallel system-event infrastructure outside `syslog`;
- replica forwarding and local writing remain covered by tests;
- adding a new `$system` event type only requires adding payload + producer logic in `syslog`.

DibbayajyotiRoy:
Closes #1485

- [ ] Docs have been added / updated (for bug fixes / features) — N/A, internal refactor, no API/config change
- [ ] CHANGELOG.md has been updated (for bug fixes / features / docs) — N/A, no user-facing change

Refactor / internal restructuring — no functional change.

### What was changed?

Unifies all `$system` event handling under `syslog` behind one generic sink, and moves every family's event-production logic into the module, as requested in #1485.

**Sink & model**
- New `SystemEventKind` routing enum carried on `SystemEvent`; it resolves the `$system/<prefix>/<instance>/<entry>` path so a single sink can route every family. It is `#[serde(skip)]`, so it never appears in the persisted record.
- `syslog/event.rs` (`SystemEvent` + `to_flat_json`), `syslog/sink.rs` (`SystemEventSink` + `LogSystemEvent`, moved out of `lifecycle`), `syslog/path.rs` (the `<prefix>/<instance>/<entry>` path + label logic that was duplicated across the two writers).
- Writers renamed to `local_writer.rs` / `forward_writer.rs` (transport only; prefix now comes from the event); `SystemLoggerBuilder` → `system_event_logger::SystemEventLoggerBuilder`.

**Production logic moved into `syslog`**
- `payload/{audit,lifecycle,log,replication,usage}.rs`
- `aggregate/{audit,usage,replication}.rs`; shared aggregation windows hoisted to `aggregate.rs` (removes the old `replication → api::audit` dependency)
- `capture/logs.rs`; deleted `api/audit.rs`; lifecycle emission stays inline.

**Kept in place per the issue**
- Configuration stays in `cfg`; `UsageCounters` stays in `storage::usage` next to its choke points; no `mod.rs` files.

**No external behavior change.** The `$system` record format and event semantics are unchanged, pinned by characterization tests added for the audit and lifecycle record key-sets and for `to_flat_json` (a test also asserts `SystemEventKind` is excluded from the serialized record).

### Related issues

Closes #1485

No. The external `$system` record format, entry paths, labels, and forwarding behavior are unchanged.

### Other information:

- Diff: 34 files, +785 / −356; the four format-defining payload files moved with a 0-line content change.
- `cargo build`, `cargo fmt --all --check`, and the full lib test suite (2136 passed / 0 failed) are green; zero new clippy warnings.
- Descoped per maintainer guidance: `payload/stats.rs`, `aggregate/lifecycle.rs`, `aggregate/counters.rs` were intentionally not created.


DibbayajyotiRoy:
@atimin please review the PR, and suggest changes and issues to resolve

reductstore/src/api/components.rs
atimin: I'd like to do deeper refactoring. We need only one component to collect all the system events.

My suggestion:

1. We change LogSystemEvent trait:

```rust
trait LogSystemEvent {
    async pub capture_audit_event(...);
    async pub capture _stat_event(..)
    async pub capture_log_event
}
```

2. if system events are disabled -  we use implementation (we don't need Option)
3. if system events are enabled, then the implementation contains all necessary aggregators. They share one global logger which is  the local or forward writer depending on the instance role 

I believe it would be better encapsulation, and in the future we will only need to extend the single trait rather than create a new component.

atimin (changes requested):

reductstore/src/api/components.rs
DibbayajyotiRoy: Done — implemented the single collector. Shape:

- New `SystemEventLogger` trait is the one component `Components` holds
  (`system_events: Arc<dyn SystemEventLogger>`), replacing `audit_logger` +
  `usage_stat_logger` + `log_capture`. Disabled instances are a no-op impl, so
  the `Option`s are gone.
- One shared `$system` writer (local on primary/standalone, forward on a
  read-only replica) built once and fanned to every family — replaces the five
  writers that were built separately before.
- `EnabledSystemEventLogger` owns all the aggregators/tasks: the audit batching
  worker, the usage timer, and the log capture. `stop()` drains them.

One thing I decided (the push/pull point from earlier): usage and logs aren't
producer-pushed the way audit is — usage is the 60s counter timer and logs are
captured passively through the global log sink — so the collector **owns and
starts those internally** instead of exposing `capture_stat_event` /
`capture_log_event`. The trait surface is:

```rust
trait SystemEventLogger {
    async fn capture_audit_event(&self, event: SystemEvent);
    fn sink(&self) -> SystemEventSink;   // shared-writer handle for lifecycle + replication
    async fn stop(&self);
}

File: reductstore/src/syslog.rs [hunk 1]
```diff
@@ -314,15 +151,148 @@ impl LogSystemEvent for DisabledSystemLogger {
...
-pub(crate) type BoxedSystemLogger = Box<dyn LogSystemEvent + Send + Sync>;
+/// The single system-event collector held by `Components`. It encapsulates
+/// every event producer (audit aggregator, usage timer, log capture) and the
+/// one shared `$system` writer they all fan through — local on a
+/// primary/standalone instance, forward on a read-only replica. A disabled
+/// instance is a no-op, so callers never need an `Option`.
+#[async_trait]
+pub(crate) trait SystemEventLogger: Send + Sync {
+    /// Capture one API audit event (batched and flushed in the background).
+    async fn capture_audit_event(&self, event: SystemEvent);
```
atimin: Do you think we could use the sink logger for all cases? This method seems a bit asymmetric compared to the other usage stats.

atimin (changes requested):
Good job! We're almost done. Please, don't forget to update the change log.

reductstore/src/syslog.rs
DibbayajyotiRoy: Good point — done. `capture_audit_event` is removed and the trait is now just
`sink()` + `stop()`.

The sink's logger is a small router: it batches `Audit`-kind events through the
existing aggregator and writes every other kind straight through the shared
writer. So all five families (audit, usage, lifecycle, replication, logs) now
emit through literally the same `SystemEventSink` — the audit batching is an
encapsulated implementation detail of the sink rather than a special method.

I kept the non-audit path synchronous on purpose: replication/lifecycle
diagnostics rely on write-then-read semantics (their tests read the record back
immediately after flush), so funneling everything through the audit channel
would have changed behavior. Added two unit tests pinning the routing (audit is
batched then flushed; other kinds are written synchronously).

CHANGELOG entry added as well.

atimin (approved):
Thank you!

PR #1497, opened 2026-07-01
Remove `include` and `exclude` query/replication API field

### LINKED ISSUE #1359: Remove deprecated query/replication API fields
## Summary
Remove deprecated query/replication API fields from the public API surface: `limit`, `each_s`, `each_t`, `include`, and `exclude`.

## Definition of done
- [ ] Remove deprecated fields from query/replication API contracts: `limit`, `each_s`, `each_t`, `include`, `exclude`
- [ ] Update request parsing/validation to reject or ignore those fields consistently (per current policy)
- [ ] Update API docs/examples to remove those fields
- [ ] Add/adjust tests covering removed-field behavior
- [ ] Verify no regressions in query and replication flows

## Links
- Release: v1.20
- Functionality: Core

vbmade2000:
Closes https://github.com/reductstore/reductstore/issues/1359

(Bug fix, feature, docs update, ...)

### What was changed?
Remove `include` and `exclude` query/replication API field

(Describe the changes)

### Related issues

(Add links to related issues)

(What changes might users need to make in their application due to this PR?)


atimin (changes requested):
Hi @vbmade2000,

It looks ready to me, but could you resolve the conflict? I also think it makes sense to remove the 'exclude' field, since it has the same logic. 

Thanks a lot for contribution!

vbmade2000:
> Hi @vbmade2000,
> 
> It looks ready to me, but could you resolve the conflict? I also think it makes sense to remove the 'exclude' field, since it has the same logic.
> 
> Thanks a lot for contribution!

Sure.

vbmade2000:
@atimin I see some tests unrelated to my changes (or at least them seem so) are failing. Is it expected? I built binary successfully and I see following tests fail. 

```
    ext::ext_repository::tests::get_ext_attachments::passes_all_attachment_keys
    ext::ext_repository::tests::get_ext_attachments::returns_error_when_attachment_limit_exceeded
    replication::replication_sender::tests::test_replication_not_found
    syslog::tests::audit_persisted_record_keyset_and_labels
```

atimin:
@vbmade2000 I think this is related. You should remove the include/exclude fields from the query structure. I think it's used for attachments. You can replace them with the "in" operator as you did before.

vbmade2000:
@atimin Should fields be removed from `QueryEntry` struct?

atimin:
> @atimin Should fields be removed from `QueryEntry` struct?

Right, you will see they're used in  reductstore/src/storage/entry/system.rs for attachment filtering.

vbmade2000:
@atimin 
I removed fields from `QueryEntry` struct but couple of tests keep failing. One is `ext::ext_repository::tests::get_ext_attachments::passes_all_attachment_keys`. 

```
assertion `left == right` failed
  left: Some(Object {"entry": Object {"$test-ext": Object {"encoding": String("ros1")}}})
 right: Some(Object {"entry": Object {"$schema": Object {"encoding": String("protobuf")}, "$test-ext": Object {"encoding": String("ros1")}}})

```

There is no error as such but looks like first write fails. Any guidance?

atimin:
@vbmade2000 could you please push your latest changes, I can't compile the current commit.

vbmade2000:
@atimin Pushed.

reductstore/src/storage/entry/system.rs
```diff
@@ -105,28 +101,12 @@ impl SystemEntryBehavior for MetaEntryBehavior {
...
                 start: Some(0),
                 stop: Some(u64::MAX),
-                include: Some(HashMap::from([("key".to_string(), key.to_string())])),
                 ..Default::default()
             })
             .await?;
         Ok(())
     }
 
-    fn apply_default_query_filters(&self, query: &mut QueryEntry) {
```
atimin, line 115: I think this is why the tests broke. This method cannot be removed.
Attachments are a special type of entry where we store metadata about the entry. They can't be removed like normal data because they must be replicated between many instances. That's why we don't remove them, but mark them for tombstoning with remove=true and then filter out them in the beginning of a query. 

Instead of removing them, replace "exclude" with "when" and add "$nin" to the "when" condition. Like you did for replications. 

Important: the $in and $nin operators, which replace the include/exclude operators, must be at the beginning of the when condition, not the end.

atimin (changes requested):

reductstore/src/storage/entry/system.rs
vbmade2000: >Instead of removing them, replace "exclude" with "when" and add "$nin" to the "when" condition. Like you did for replications.

Do you mean replace in `apply_default_query_filters` method itself?

vbmade2000: This method `apply_default_query_filters` uses `include` and `exclude` fields from `QueryEntry` struct. If we remove those fields from the struct as you suggested then there is no use of that method.

atimin: Hi @vbmade2000 , sorry, I haven't use include\exclude parameters ages and forgot how they really work.


My suggestion for the filter:

```rust
    fn apply_default_query_filters(&self, query: &mut QueryEntry) {
        let exclude = json!([{
            "$and": [
                {"$has": "remove"}, {"&remove": {"$eq": true}}
            ]
        }]);
        let when = query.when.get_or_insert_with(|| json!({}));
        match when.as_object_mut().unwrap().entry("$not") {
            Vacant(entry) => {
                entry.insert(exclude);
            }
            Occupied(mut entry) => {
                entry.get_mut().as_array_mut().unwrap().push(exclude.get(0).unwrap().clone());
            }
        }
```
here we don't check include, meaning we can't get the removed entries explicitly anymore. It's fine.


I was talking about $in and $nin operators, it was wrong too. To migrate from include/exclude:


include={"lablel1":"a", "label2", "b"}
 => when = {
 ...
 "$and": [
   {"$has": "label1", "&label1": { "$eq":  "a"}, 
   {"$has": "label2", "&label2": {"$eq": "b"}}
 ] 
}

exclude={"lablel1":"a", "label2", "b"}
 => when = {
... 
"$not": [
   {"$or": [
     {"$has": "label1", "&label1": { "$eq":  "a"}, 
     {"$has": "label2", "&label2": {"$eq": "b"}}
  ] }
 ]
}
 

I apologize for the confusion.

vbmade2000: That's fine. I'll look into it. Thanks!

vbmade2000: Can you please check my latest commit at https://github.com/reductstore/reductstore/pull/1497/changes/bcdf937d75ec3c536169fa32315f9111f45e890b and tell me if changes are good? Because those two tests are still failing.

atimin: Good that we have tests.. 

We need to update the prepare_write method:

```rust
#[async_trait]
impl SystemEntryBehavior for MetaEntryBehavior {
    async fn prepare_write(&self, entry: &Entry, labels: &Labels) -> Result<(), ReductError> {
        let Some(key) = labels.get("key") else {
            return Err(unprocessable_entity!(
                "System entry '{}' records must contain label 'key'",
                entry.name()
            ));
        };

        if labels.get("remove").is_some_and(|value| value == "true") {
            return Err(unprocessable_entity!(
                "System entry '{}' does not support writing records with label 'remove=true'; use record update",
                entry.name()
            ));
        }

        // escape system names started with $
        let key = if key.starts_with("$") {
            format!("${}", key)
        } else {
            key.to_string()
        };

        // remove previous version
        let _ = entry
            .query_remove_records(QueryEntry {
                start: Some(0),
                stop: Some(u64::MAX),
                when: Some(json!({
                        "$has": "key",
                        "&key": {"$eq": key}
                })),
                ..Default::default()
            })
            .await?;
        Ok(())
    }
```

You remove the include parameter, but we still have to filter attachments by key to remove the previous version before writing a new one.

vbmade2000: Great. Tests are passing now. Thanks for suggestion.

vbmade2000:
@atimin I think it is ready for review now.

atimin (approved):
Perfect, thank you! I think we have only each_n to remove. Let's do it in another PR.

PR #1506, opened 2026-07-03
feat: support query entry glob patterns (#1505)

### LINKED ISSUE #1505: Feature: extend I/O run_query entry filters with recursive wildcards and exclusions
## Summary

Extend `POST /api/v1/io/:bucket_name/q` entry selection so `entries` can express path-aware include and exclude rules instead of only exact names or trailing-prefix wildcards.

## Current behavior

Today the implementation only supports:

- exact entry names
- `"*"` for all entries
- trailing `*` as a prefix wildcard

The current docs describe only exact names and prefix wildcards:

- https://www.reduct.store/docs/http-api/io-api/run_query
- `["*"]` selects all entries
- `["prefix-*"]` selects all entries with names starting with `prefix-`

The current matcher in `reductstore/src/storage/bucket/query.rs` only checks exact equality or `strip_suffix('*')` prefix matching.

## Requested behavior

Support richer `entries` patterns for the I/O query API:

1. `*` keeps its current meaning as the simple wildcard within a single path segment.
2. `**` becomes a recursive wildcard across path segments.
3. `!` at the start of a pattern excludes matching entries.

## Matching rules

- `/a/*/b` matches `/a/x/b`
- `/a/*/b` does **not** match `/a/x/d/b`
- `/a/**/b` matches `/a/x/b`
- `/a/**/b` matches `/a/x/d/b`
- recursive patterns like `/**/**/` should be accepted
- `!` exclusions should use the same matching rules as positive patterns
- exact names must continue to work unchanged
- existing prefix-style `*` behavior must remain backward compatible

## Example

```json
{
  "query_type": "QUERY",
  "entries": [
    "/a/**/b",
    "!/a/private/**",
    "cam-*"
  ]
}
```

Expected result:

- include any entry matching `/a/**/b`
- exclude anything under `/a/private/**`
- keep existing `cam-*` prefix matching working as before

## Why this matters

Entry names are often used like paths. The current prefix-only wildcard is too limited when data is organized hierarchically and clients need to:

- select nested branches without enumerating every level
- exclude subtrees from a broad query
- mix legacy flat naming patterns with hierarchical path patterns

## Suggested acceptance criteria

- `entries` supports exact, include, and exclude patterns
- `*` matches within one segment only
- `**` matches across multiple segments
- `!pattern` excludes matches after includes are resolved
- existing exact-name and trailing-prefix `*` behavior remains compatible
- docs at https://www.reduct.store/docs/http-api/io-api/run_query are updated with examples and precedence rules
- tests cover include-only, exclude-only, mixed include/exclude, `*`, `**`, and backward-compatibility cases

gitcommit90:
Closes #1505

Feature

### What was changed?

- Adds path-aware query entry pattern matching for `*` single-segment wildcards and `**` recursive wildcards.
- Preserves existing exact-name matching and legacy prefix-style patterns like `cam-*`.
- Adds `!` exclusion patterns that are applied after includes, including exclusion-only filtering.
- Keeps wildcard pattern matches restricted to entries queryable by wildcard.

### Related issues

- #1505

No. Existing exact entry filters, `*`, and prefix-style wildcard behavior are preserved.

### Other information:

Validation run locally in Docker/Rust 1.91:

```console
cargo fmt --all
cargo test -p reductstore storage::bucket::query --features "fs-backend web-console"
# 13 passed; 0 failed

cargo check --workspace --features "fs-backend web-console"
# Finished dev profile
```

### DIFF
--- a/reductstore/src/storage/bucket/query.rs
+++ b/reductstore/src/storage/bucket/query.rs
@@ -27,6 +27,69 @@ pub(super) struct MultiEntryQuery {
     last_access: Instant,
 }
 
+fn entry_matches_pattern(entry: &str, pattern: &str) -> bool {
+    let pattern = pattern.trim_start_matches('/');
+
+    if !pattern.contains('*') {
+        return entry == pattern;
+    }
+
+    let entry_parts: Vec<&str> = entry.split('/').collect();
+    let pattern_parts: Vec<&str> = pattern.split('/').collect();
+
+    fn segment_matches(entry: &str, pattern: &str) -> bool {
+        if pattern == "**" {
+            return true;
+        }
+
+        let mut rest = entry;
+        let mut parts = pattern.split('*').peekable();
+
+        if let Some(first) = parts.next() {
+            if !first.is_empty() {
+                let Some(stripped) = rest.strip_prefix(first) else {
+                    return false;
+                };
+                rest = stripped;
+            }
+        }
+
+        while let Some(part) = parts.next() {
+            if part.is_empty() {
+                continue;
+            }
+
+            if parts.peek().is_none() {
+                return rest.ends_with(part);
+            }
+
+            let Some(index) = rest.find(part) else {
+                return false;
+            };
+            rest = &rest[index + part.len()..];
+        }
+
+        pattern.ends_with('*') || rest.is_empty()
+    }
+
+    fn matches_from(entry_parts: &[&str], pattern_parts: &[&str]) -> bool {
+        match pattern_parts.split_first() {
+            None => entry_parts.is_empty(),
+            Some((&"**", tail)) => {
+                matches_from(entry_parts, tail)
+                    || (!entry_parts.is_empty() && matches_from(&entry_parts[1..], pattern_parts))
+            }
+            Some((pattern, tail)) => {
+                !entry_parts.is_empty()
+                    && segment_matches(entry_parts[0], pattern)
+                    && matches_from(&entry_parts[1..], tail)
+            }
+        }
+    }
+
+    matches_from(&entry_parts, &pattern_parts)
+}
+
 impl Bucket {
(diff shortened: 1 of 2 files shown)


reductstore/src/storage/bucket/query.rs
```diff
@@ -422,6 +504,82 @@ mod tests {
         );
     }
 
+    #[rstest]
+    #[tokio::test]
+    async fn filters_by_single_segment_wildcard(#[future] bucket: Arc<Bucket>) {
```
atimin: since you moved the matching into a separate method, it makes sense to make tests shallow and test only matching logic without writing real records.

atimin: you can  also use the #case macros to reduce duplication in tests.

reductstore/src/storage/bucket/query.rs
```diff
@@ -27,6 +27,69 @@ pub(super) struct MultiEntryQuery {
...
 }
 
+fn entry_matches_pattern(entry: &str, pattern: &str) -> bool {
+    let pattern = pattern.trim_start_matches('/');
+
+    if !pattern.contains('*') {
+        return entry == pattern;
+    }
+
+    let entry_parts: Vec<&str> = entry.split('/').collect();
```
atimin, line 43: I think we should check if it's a simple wildcard like 'cam-*' before doing segmentation. I suspect your code would be backward compatible with 'cam-1/something-else'.

atimin (changes requested):
Hi @gitcommit90,
Thank you for your contribution.

The PR is almost perfect, but there are a few issues to address. Please read my comments.

Don't forget to update the changelog and add "by @gitcommit90" so you will be listed as a contributor. 

Thanks!

atimin (approved):

atimin:
Perfect! Thank you!

PR #1525, opened 2026-07-07
Check data folder free disk space before writing records

### LINKED ISSUE #1522: Check data-folder free disk space before writing records
## Problem
Before writing a record, ReductStore checks quota but can still proceed when the data folder has insufficient free disk space.

## Proposed solution
Before accepting a record write, check both:
- configured quota
- available free disk space on the filesystem containing the data folder

Reject the write early when either limit cannot fit the incoming record.

## Definition of done
- Record writes fail before writing when the data folder filesystem lacks enough free space.
- Existing quota behavior is preserved.
- Add coverage for the free-space failure path.

upuddu:
Closes #1522

Bug fix

### What was changed?

Before accepting a record write, ReductStore checked the configured quota but could still proceed when the data folder's filesystem had run out of free space. This adds a free-space check to the write path:

- `Bucket::begin_write` now rejects a write with `507 Insufficient Storage` when the incoming record doesn't fit in the space available on the filesystem holding the data folder. The check runs right after the existing quota check, so FIFO eviction still gets a chance to free space first.
- Free space is queried with `fs4::available_space` for cross-platform support (the in-tree `nix`/`libc` are Unix-only, and CI covers Windows).
- Existing quota behavior (NONE / FIFO / HARD) is unchanged.

### Related issues

#1522

No. Writes that previously succeeded still succeed; only writes that wouldn't fit on disk are now rejected up front instead of failing mid-write.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Reject record writes before writing when the data folder filesystem has insufficient free disk space, in addition to the existing quota check, [PR-1522](https://github.com/reductstore/reductstore/issues/1522)
 - Support glob-like entry patterns in replication filters, including path-aware `*` and `**` wildcards and exclusion patterns, [PR-1507](https://github.com/reductstore/reductstore/issues/1507) by @rohankumardubey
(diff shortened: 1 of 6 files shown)


atimin:
Hi @upuddu , thank you for you contribution! The patch looks great to me. Please, resolve  the conflict and I'll merge it.

atimin:
Never mind, @upuddu, I found out how to do it myself! Thanks!

PR #1527, opened 2026-07-07
Pipeline replication batch sending (#1061, supersedes #1415)

### LINKED ISSUE #1061: Feature: Optimize replication task
### Current problem

Currently, the task processes transaction logs in one thread and does not parallelise writes to different entries. 

### Possible solution

We already use a token runtime in the remote bucket; let's apply the same approach at the task sender level and spawn a Tokio task for each entry. 

### Additional context

_No response_

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

DibbayajyotiRoy:
Closes #1061. Supersedes #1415 by @FirasCh3, whose commits and approach this builds on.

## What this does
Overlaps preparing the next entry's batch with sending the current one: one `RemoteBucket`, one HTTP send in flight, per your call on the single-client design.

## Review remarks from #1415, addressed
1. **Duplicated result handling** unified into a single `drain_pending` helper used by both the in-loop and final-drain sites.
2. **Transaction acks moved into the send task**: `pop_front` happens inside the spawned task on success, so the task returns only the bucket, counter results, and the active flag.
3. **No `.expect` on the join handle**: a panicked send task is logged and the bucket is rebuilt from the factory, so an internal bug degrades one pass instead of crashing the database. Every early return (including the `BrokenLog` path) drains the pending send first, so the bucket can no longer be lost in flight.
4. **CHANGELOG** credits #1415.

## Tests
`test_pipelined_multi_entry_drains_final_send`, `test_pipelined_ack_is_per_entry` (two entries, 429 on one: failing entry retains transactions, succeeding entry pops), `test_broken_log_drains_pending_send`, `test_send_task_panic_recovers_bucket` (a subsequent run works after a panic). All 235 replication tests green; fmt and clippy clean.


DibbayajyotiRoy:
@atimin please give a review

atimin (changes requested):
Hi @DibbayajyotiRoy ,

Thank you very much for your contribution. Unfortunately, I think the implementation is a bit overcomplicated. We can't use a factory for the remote bucket because it conflicts with the current design. It has an internal state, which allows it to recover after communication errors (see state pattern). 

I suggest addressing my change request to the previous PR since the optimization is complete and we only need to elaborate on the implementation. 

Thanks again,

reductstore/src/replication/replication_sender.rs
```diff
@@ -60,142 +89,240 @@ impl ReplicationSender {
...
         for entry_name in entries.iter() {
-            let log = {
-                // Take only the handle, drop the map lock before touching the log itself.
-                let map = self.log_map.read().await?;
-                match map.get(entry_name) {
-                    Some(log) => Arc::clone(log),
-                    None => continue, // log might be removed
+            // Depth-1 pipelining: send the previously prepared batch while
+            // preparing this entry's batch.
+            let (send_result, prepare_result) = tokio::join!(
```
atimin: i would do it more obviously:

``` rust
let mut send_task = None;

if let Some(handle) = send_task {
    if let Error(err) = handle.await {
      // print error.  
   }
}

let batch = prepare_batch(....);
send_task = Some(send_batch(batch, ...);
```

the order and logic are clear in this case.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Pipelined replication batch sending to overlap preparing the next batch with sending the current one, [PR-NNNN](https://github.com/reductstore/reductstore/pulls) (based on [PR-1415](https://github.com/reductstore/reductstore/pull/1415) by @FirasCh3)
```
atimin: add yourself too.

atimin (changes requested):
Hi @DibbayajyotiRoy , we're much closer =D

CHANGELOG.md
atimin: and use please the ID not NNNN

reductstore/src/replication/replication_sender.rs
DibbayajyotiRoy: done

atimin (approved):
Perfect! Thank you!

PR #1538, opened 2026-07-10
Replication: add optional payload compression (#1348)

### LINKED ISSUE #1348: Replication: add optional payload compression
## Problem
Replication transfers batch payloads without wire compression, which increases bandwidth usage on constrained links.

## Proposed solution
Add a new replication task option `compression` to enable payload compression during transfer.

## Acceptance criteria
- Replication task supports `compression` option with values: `none`, `zstd`, `gzip`.
- Default is `none` for backward compatibility.
- Compression is applied to batch payload transfer only.
- Replication automatically skips compression for already compressed MIME types.
- Receiver correctly handles compressed payloads and preserves existing replication behavior.
- Metrics include compressed bytes and compression ratio.

DibbayajyotiRoy:
Closes #1348

## What

Adds a per-replication-task `compression` setting (`none` | `zstd` | `gzip`, default `none`) that compresses replication batch payloads during transfer using **standard HTTP `Content-Encoding`** (reworked per review feedback — the original per-record scheme is gone).

## Design

- **Sender** (`client_wrapper.rs`): when compression is enabled, the batch body is compressed as a whole and sent with `Content-Encoding: gzip|zstd`. Note on the reqwest suggestion: `ClientBuilder::gzip()/zstd()` only auto-decompress *responses* ("The request body is not automatically compressed" per the docs), so the request body is compressed explicitly. The raw payload size is carried in the already-supported `x-reduct-content-length` header, so the receiver's `content-length == Σ record lengths` framing check works unchanged after decompression.
- **Receiver**: `tower-http`'s `RequestDecompressionLayer` (gzip + zstd features) on the router decompresses request bodies transparently — no handler changes; this also makes `Content-Encoding` work for regular client writes, not just replication.
- **Compatibility**: default is `none`; the setting is `#[serde(default)]` and proto field 14 defaults to `NONE`, so existing configs/clients are unaffected. The sender checks the destination's `x-reduct-api` version and falls back to uncompressed (with a warning) for servers older than 1.21, which can't decompress requests.
- **Provisioning**: `RS_REPLICATION_<id>_COMPRESSION=none|zstd|gzip`.
- Per review, the MIME-type skip list and compressed-bytes/ratio metrics were dropped.

## Tests

- Unit: sender body compression + header rewrite (gzip/zstd round-trip, error propagation, version gating), router-level receiver tests posting real gzip/zstd batches through the decompression layer (mixed records, labels, corrupted body → 400), settings CRUD/proto round-trip.
- Integration: replication API tests cover the `compression` setting; the CI replication test matrix now runs `none`, `zstd`, and `gzip` end-to-end between two instances.
- `cargo fmt --check` clean; full `cargo test`: 2241 passed, 0 failed.

## Out of scope

Console UI and client SDKs (separate repos) need the `compression` field added to their replication settings models; serde defaults keep them working unchanged meanwhile.


DibbayajyotiRoy:
@atimin please have a review

atimin (changes requested):
Hi @DibbayajyotiRoy ,

Thank you very much for your PR. Unfortunately, it's not the way I would go. The HTTP protocol natively supports compression; we only need to enable it: https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.gzip.

I apologize for not mentioning this in the issue; it was obvious to me.  I see that some of the requirements would be difficult to implement with automatic compression. However, we can sacrifice some of them, such as:

* Replication automatically skips compression for already compressed MIME types.
* Metrics include compressed bytes and compression ratio.

Thank you!

DibbayajyotiRoy:
> Hi @DibbayajyotiRoy ,
> 
> Thank you very much for your PR. Unfortunately, it's not the way I would go. The HTTP protocol natively supports compression; we only need to enable it: https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.gzip.
> 
> I apologize for not mentioning this in the issue; it was obvious to me. I see that some of the requirements would be difficult to implement with automatic compression. However, we can sacrifice some of them, such as:
> 
> * Replication automatically skips compression for already compressed MIME types.
> * Metrics include compressed bytes and compression ratio.
> 
> Thank you!


On it!!

DibbayajyotiRoy:
Hi @atimin, thanks for the direction — reworked the PR to use standard HTTP `Content-Encoding` and dropped the per-record scheme, the MIME skip list, and the metrics as you suggested.

One detail worth flagging: reqwest's `ClientBuilder::gzip()/zstd()` only handles *response* decompression — per the docs, "The request body is not automatically compressed." Since replication pushes data (POST), the sender compresses the batch body explicitly and sets `Content-Encoding`, and the server side gets a `tower-http` `RequestDecompressionLayer` so any compressed request body (replication or regular client writes) is decompressed transparently.

Two small things I kept, let me know if you'd rather drop them too:
- the raw batch size travels in the existing `x-reduct-content-length` header, so the receiver's content-length-vs-record-sum check still works after decompression;
- the sender checks the destination's `x-reduct-api` version and falls back to uncompressed for servers older than 1.21 (they can't decompress request bodies).

The diff is now much smaller. CI replication tests run the zstd and gzip paths end-to-end.

reductstore/src/api/http/entry/write_batched.rs
```diff
@@ -789,6 +789,118 @@ mod tests {
         assert!(err.message().contains("ingress bytes"));
     }
 
+    mod compressed {
```
atimin: I don't think we need  these tests since the backend support compressed

reductstore/src/replication/remote_bucket/client_wrapper.rs
```diff
@@ -169,6 +198,52 @@ impl BucketWrapper {
...
+    /// is set from the compressed body by the HTTP client.
+    async fn compress_body(
+        headers: &mut HeaderMap,
+        body: impl Stream<Item = Result<Bytes, ReductError>>,
+        compression: ReplicationCompression,
+    ) -> Result<Vec<u8>, ReductError> {
+        let mut raw = Vec::new();
+        let mut body = std::pin::pin!(body);
+        while let Some(chunk) = body.next().await {
+            raw.extend_from_slice(&chunk?);
```
atimin: Gzip and ZSTD can compress data in a stream. Instead of receiving the entire stream in memory, compression can be performed in chunks as the stream is received.

reductstore/src/replication/remote_bucket/client_wrapper.rs
```diff
@@ -169,6 +198,52 @@ impl BucketWrapper {
...
+                    .write_all(&raw)
+                    .and_then(|_| encoder.finish())
+                    .map(|compressed| ("gzip", compressed))
+                    .map_err(|err| {
+                        internal_server_error!("Failed to compress batch payload: {}", err)
+                    })?
+            }
+        };
+
+        let raw_content_length = headers
```
atimin: I didn't understand this. We use x-reduct-content-length as a workaround for the JavaScript Fetch API, which doesn't allow you to work with streams and set content-length, so they are always the same.

reductstore/src/replication/remote_bucket/client_wrapper.rs
```diff
@@ -252,12 +339,14 @@ impl ReductBucketApi for BucketWrapper {
...
 
-        let response = request
-            .headers(headers)
-            .body(Body::wrap_stream(stream))
-            .send()
-            .await;
+        let request = if self.compression == ReplicationCompression::None {
+            request.headers(headers).body(Body::wrap_stream(stream))
+        } else {
+            let compressed = Self::compress_body(&mut headers, stream, self.compression).await?;
```
atimin: we need streams , as I described above.

atimin (changes requested):
The direction is correct now, but still I see some issues. 

Please, merge the latest main since we made some chnages in replications and I'd like to test everything together.

DibbayajyotiRoy:
Thanks for the review @atimin — all three points addressed in 49226a8d, and the latest main (including the auto-create destination bucket change) is merged in:

- **Streaming compression**: the batch body is now compressed as a stream (`async-compression` Gzip/Zstd encoders wrapped around the record chunk stream) — nothing is buffered in memory anymore. `async-compression` was already in the dependency tree via `tower-http`.
- **`x-reduct-content-length`**: dropped — I had repurposed it because a streamed compressed body has no `content-length` (chunked) and the decompression layer consumes the header on the receiving side. Instead, the receiver now derives the expected payload size from the sum of the per-record lengths when no `content-length` header is present, so the header keeps its original meaning.
- **Batch handler compression tests**: removed; kept only the sender-side unit tests (stream round-trip, error propagation, version gating).

The compression version gating is also applied on the new `create_bucket` path from #1539.

atimin (approved):
Now it's perfect!

Thank you very much for the great contribution 🙏

PR #1539, opened 2026-07-11
auto-create missing replication destination buckets

### LINKED ISSUE #1537: Replication should create missing destination bucket before failing
## Summary

A replication task should try to create the configured destination bucket on the remote ReductStore instance when the bucket does not exist, before marking the remote bucket unavailable or failing replication.

## Current behavior

The replication remote-bucket state probes the destination with `GET /api/v1/b/{dst_bucket}`. If that returns an error, including `404 Not Found`, the task moves to the unavailable state and keeps retrying the bucket lookup.

That means a valid replication setup fails until an operator manually creates the destination bucket.

## Desired behavior

When the destination bucket lookup returns `404 Not Found`:

1. Try to create `settings.dst_bucket` on the destination ReductStore instance.
2. If creation succeeds, continue replication into the newly created bucket.
3. If creation returns conflict/already-exists, re-fetch the bucket and continue if it is now available.
4. If creation is denied or fails for another reason, preserve the current error reporting/retry behavior.

For non-404 errors such as connection failures, authentication failures, permission failures on bucket lookup, or TLS/client errors, keep the current unavailable/error path.

## Why

Replication is often configured from the source side. Requiring operators to pre-create each destination bucket is easy to miss and makes first-run replication brittle, especially for edge-to-cloud deployments and provisioning-based setups.

Auto-creating only on a missing destination bucket keeps the existing failure behavior for real connectivity/auth/configuration problems while making the common bootstrap path smoother.

## Acceptance criteria

- Replication attempts to create the remote destination bucket only when `GET dst_bucket` returns `404 Not Found`.
- After successful creation, the replication task writes pending records without requiring a restart.
- A concurrent creation/conflict race is handled gracefully by re-fetching the bucket.
- Permission/auth/network errors are not hidden as successful bucket creation.
- Unit tests cover initial state and unavailable-state retry behavior for:
  - bucket exists,
  - bucket missing then created,
  - bucket missing and creation fails,
  - bucket created concurrently.
- Integration or API-level test coverage verifies a replication task can start with an absent destination bucket and deliver records after creating it.

rohankumardubey:
Closes #1537

- [x] Tests for the changes have been added
- [ ] Docs have been added / updated (not required)
- [ ] CHANGELOG.md has been updated (will be added after the PR number is available)

Bug fix

### What was changed?

Replication now creates a missing destination bucket before failing.

When the destination bucket lookup returns `404 Not Found`, ReductStore creates the bucket using only its name and continues replication. If another process creates the bucket at the same time, a `409 Conflict` triggers a retry of the bucket lookup.

Added unit coverage for successful creation, concurrent creation, and creation failures. Updated the replication test setup so the destination bucket is not pre-created.

### Related issues

https://github.com/reductstore/reductstore/issues/1537

No.

### Other information:

The destination bucket is created without copying source bucket settings. Users who need specific bucket settings should provision the destination bucket in advance.

### DIFF
--- a/reductstore/src/replication/remote_bucket/states.rs
+++ b/reductstore/src/replication/remote_bucket/states.rs
@@ -5,12 +5,32 @@ mod bucket_available;
 mod bucket_unavailable;
 mod initial_state;
 
+use crate::replication::remote_bucket::client_wrapper::{BoxedBucketApi, BoxedClientApi};
 use crate::replication::remote_bucket::ErrorRecordMap;
 use crate::replication::Transaction;
 use async_trait::async_trait;
 pub(super) use initial_state::InitialState;
-use reduct_base::error::ReductError;
+use reduct_base::error::{ErrorCode, ReductError};
 use reduct_base::io::BoxedReadRecord;
+
+async fn get_or_create_bucket(
+    client: &BoxedClientApi,
+    bucket_name: &str,
+) -> Result<BoxedBucketApi, ReductError> {
+    match client.get_bucket(bucket_name).await {
+        Ok(bucket) => Ok(bucket),
+        Err(err) if err.status() == ErrorCode::NotFound => {
+            match client.create_bucket(bucket_name).await {
+                Ok(bucket) => Ok(bucket),
+                Err(err) if err.status() == ErrorCode::Conflict => {
+                    client.get_bucket(bucket_name).await
+                }
+                Err(err) => Err(err),
+            }
+        }
+        Err(err) => Err(err),
+    }
+}
 
 /// A state of the remote bucket.
 
--- a/reductstore/src/replication/remote_bucket/states.rs
+++ b/reductstore/src/replication/remote_bucket/states.rs
@@ -5,32 +5,13 @@ mod bucket_available;
 mod bucket_unavailable;
 mod initial_state;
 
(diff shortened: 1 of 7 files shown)


rohankumardubey:
@atimin This is ready for review.

reductstore/src/replication/remote_bucket/states.rs
```diff
@@ -5,13 +5,33 @@ mod bucket_available;
...
+use crate::replication::remote_bucket::client_wrapper::{BoxedBucketApi, BoxedClientApi};
 use crate::replication::remote_bucket::ErrorRecordMap;
 use crate::replication::Transaction;
 use async_trait::async_trait;
 pub(super) use initial_state::InitialState;
-use reduct_base::error::ReductError;
+use reduct_base::error::{ErrorCode, ReductError};
 use reduct_base::io::BoxedReadRecord;
 
+async fn get_or_create_bucket(
```
atimin: I would move to the ClientWrapper trait as a method with default implementation.

atimin (changes requested):
Hi @rohankumardubey , 

It's perfect. Only consider moving the get_or_create method into the trait. I think that's a more idiomatic way to extend the functionality. 

Thank you!

atimin (approved):
Perfect! Thank you for the next part of your excellent contribution.

Issue #1288, opened 2026-04-03
bug(audit): client_ip behind AWS ALB resolves to proxy private IP instead of X-Forwarded-For


mother-6000:
## Summary
When ReductStore is deployed behind AWS ALB, audit records store the private proxy IP (e.g. `10.0.x.x`) instead of the real client IP from `X-Forwarded-For`.

## Environment
- Deployment: ReductStore behind AWS Application Load Balancer (ALB)
- Observed in: audit records (`client_ip`)

## Steps to Reproduce
1. Deploy ReductStore behind AWS ALB.
2. Send requests from a public client IP through ALB.
3. Inspect audit records in `$audit` bucket.

## Actual Result
`client_ip` is recorded as ALB/private network address (e.g. `10.0.x.x`).

## Expected Result
`client_ip` should be resolved from forwarded headers (`X-Forwarded-For` / `Forwarded`) and contain the original client public IP.

## Suspected Root Cause
In `reductstore/src/api/http/middleware/client_ip.rs`, forwarded headers are only trusted when peer IP is loopback:

```rust
fn is_trusted_proxy(ip: IpAddr) -> bool {
    ip.is_loopback()
}
```

Behind ALB, peer IP is typically private (`10.x.x.x`), not loopback, so the code falls back to peer IP and ignores forwarded headers.

## Suggested Direction
Make trusted proxy behavior configurable (CIDR allowlist and/or trusted hop count), so common reverse-proxy setups (ALB, ingress, etc.) can safely use forwarded headers.

PR #1546, opened 2026-07-13
fix(audit): resolve client_ip from multi-hop RFC 7239 Forwarded header

### LINKED ISSUE #1545: parse_forwarded_for drops multi-hop RFC 7239 Forwarded headers
### Summary

`parse_forwarded_for` in `reductstore/src/api/http/middleware/client_ip.rs`
only splits the `Forwarded` header on `;`. RFC 7239 separates multiple proxy
hops with `,`. When a request carries a standards-compliant multi-hop
`Forwarded` header, the parser fails and returns `None`.

Tested against `main` @ `2f9daa5`.

### What happens

For a header like:

    Forwarded: for=192.0.2.43, for=198.51.100.17

the function splits only on `;`, so the first `for=` value becomes the entire
remaining string `192.0.2.43, for=198.51.100.17`. That fails `parse::<IpAddr>()`,
and the function returns `None`.

### Expected

Per RFC 7239 the client is the first hop, so this should resolve to
`192.0.2.43`, the same way `parse_x_forwarded_for` already takes the first
comma-separated element.

### Impact

This matters when ReductStore is behind a trusted proxy and a token has an IP
allowlist. With the parse returning `None` (and no `X-Forwarded-For` present),
`client_ip_from_request` falls back to the proxy's own IP via
`unwrap_or(peer_ip)`. That proxy IP is stored in `x-reduct-client-ip` and used
for the allowlist check, so a token restricted to the real client's IP is
evaluated against the proxy's IP and the request is rejected with
`Token is not allowed for client IP <proxy-ip>`. A legitimate client is locked
out. This fails closed, so it is a correctness/availability problem, not an auth
bypass.

In practice most proxies also send `X-Forwarded-For`, and the code falls
through to it, so this only bites when a proxy sends a standards-compliant
`Forwarded` header without `X-Forwarded-For`.

### Precedent

This is the same failure mode as #1288 (client IP resolving to the proxy behind
AWS ALB), which was fixed for `X-Forwarded-For` in #1296. The `Forwarded` header
path still has the same problem, because `parse_forwarded_for` never splits on
`,`.

### Root cause

`parse_forwarded_for` iterates `value.split(';')` and never splits on `,`. Its
sibling `parse_x_forwarded_for` does split on `,` and takes `.next()`, so the
two header parsers are inconsistent and only the `Forwarded` one is affected.

### Reproduction

Adding these cases to the existing `parses_forwarded_header` rstest in
`middleware.rs`:

    #[case("for=192.0.2.60;proto=http", Some("192.0.2.60"))]           // passes today
    #[case("proto=http;for=192.0.2.60", Some("192.0.2.60"))]           // passes today
    #[case("for=192.0.2.43, for=198.51.100.17", Some("192.0.2.43"))]   // returns None
    #[case("for=192.0.2.43, for=198.51.100.17;proto=http", Some("192.0.2.43"))] // returns None

The last two fail:

    left: None
    right: Some("192.0.2.43")

### Proposed fix

Split the value on `,` first, then run the current `;`-based parameter parse on
each hop, returning the first hop that yields a valid `for=` IP. The five
existing test cases contain no comma, so their behavior does not change.

### Note

While reading this code I also noticed `for="[2001:db8::1]:8080"` (quoted
bracketed IPv6 with a port) returns `None`. That looks like a separate gap; I
have left it out of this report to keep it to one bug.

I would be happy to fix this. Could you assign it to me?

LordAizen1:
Fixes #1545.

`parse_forwarded_for` only splits the `Forwarded` header on `;`, never on `,`.
RFC 7239 uses commas between proxy hops, so something like

    Forwarded: for=192.0.2.43, for=198.51.100.17

ends up shoving `192.0.2.43, for=198.51.100.17` into the first `for=` value.
That doesn't parse as an IP, so the function just returns `None`.

The annoying part is what happens next. Behind a trusted proxy that sends
`Forwarded` but no `X-Forwarded-For`, `client_ip_from_request` falls back to the
proxy's own IP. So if a token has an IP allowlist, it gets checked against the
proxy IP instead of the real client and the request is rejected with
`Token is not allowed for client IP <proxy-ip>`. It fails closed, so it's not a
bypass, just a valid client getting locked out. Same thing as #1288, which was
fixed for `X-Forwarded-For` in #1296. The `Forwarded` path never got the same
treatment.

To be fair this is fairly narrow. Most proxies send `X-Forwarded-For` too and
the code falls through to that, so you only hit it with a spec-compliant
`Forwarded` header and no XFF.

The fix splits on `,` first and runs the existing per-hop parsing on each piece,
taking the first hop with a real `for=` IP. That's the same thing
`parse_x_forwarded_for` already does with its first element. I just pulled the
per-hop logic into a helper, didn't change it.

Added two cases to the existing `parses_forwarded_header` rstest for the
multi-hop header (one with a trailing `;proto=http` on the second hop). The old
single-hop cases have no comma, so nothing changes for them.

### DIFF
--- a/reductstore/src/api/http/middleware.rs
+++ b/reductstore/src/api/http/middleware.rs
@@ -220,6 +220,7 @@ mod tests {
     #[case("for=_hidden", None)]
     #[case("for=192.0.2.43, for=198.51.100.17", Some("192.0.2.43"))]
     #[case("for=192.0.2.43, for=198.51.100.17;proto=http", Some("192.0.2.43"))]
+    #[case("for=_hidden, for=198.51.100.17;proto=http", Some("198.51.100.17"))]
     fn parses_forwarded_header(#[case] input: &str, #[case] expected: Option<&str>) {
         assert_eq!(
             parse_forwarded_for(input).map(|ip| ip.to_string()),
--- a/reductstore/src/api/http/middleware.rs
+++ b/reductstore/src/api/http/middleware.rs
@@ -218,6 +218,8 @@ mod tests {
     #[case("for=\"[2001:db8:cafe::17]\"", Some("2001:db8:cafe::17"))]
     #[case("by=203.0.113.60;proto=http", None)]
     #[case("for=_hidden", None)]
+    #[case("for=192.0.2.43, for=198.51.100.17", Some("192.0.2.43"))]
+    #[case("for=192.0.2.43, for=198.51.100.17;proto=http", Some("192.0.2.43"))]
     fn parses_forwarded_header(#[case] input: &str, #[case] expected: Option<&str>) {
         assert_eq!(
             parse_forwarded_for(input).map(|ip| ip.to_string()),
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Resolve client IP from multi-hop RFC 7239 `Forwarded` headers so token IP allowlists compare against the real client instead of the proxy, [PR-1546](https://github.com/reductstore/reductstore/pull/1546) by @LordAizen1
 - Normalize bucket history timestamps when only attachment metadata entries contain records, [PR-1534](https://github.com/reductstore/reductstore/pull/1534)
 
 ## 1.20.9 - 2026-07-08
--- a/README.md
+++ b/README.md
@@ -164,6 +164,7 @@ Thanks to everyone who has contributed to ReductStore.
   <a href="https://github.com/victor1234"><img src="https://avatars.githubusercontent.com/u/1102205?v=4" width="48" height="48" alt="@victor1234" /></a>
   <a href="https://github.com/aschenbecherwespe"><img src="https://avatars.githubusercontent.com/u/94011659?v=4" width="48" height="48" alt="@aschenbecherwespe" /></a>
   <a href="https://github.com/renghen"><img src="https://avatars.githubusercontent.com/u/271285?v=4" width="48" height="48" alt="@renghen" /></a>
+  <a href="https://github.com/LordAizen1"><img src="https://avatars.githubusercontent.com/u/119096690?v=4" width="48" height="48" alt="@LordAizen1" /></a>
 </p>
 
 Your support fuels our passion and drives us to keep improving.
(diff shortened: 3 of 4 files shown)


reductstore/src/api/http/middleware.rs
```diff
@@ -218,6 +218,8 @@ mod tests {
     #[case("for=\"[2001:db8:cafe::17]\"", Some("2001:db8:cafe::17"))]
     #[case("by=203.0.113.60;proto=http", None)]
     #[case("for=_hidden", None)]
+    #[case("for=192.0.2.43, for=198.51.100.17", Some("192.0.2.43"))]
+    #[case("for=192.0.2.43, for=198.51.100.17;proto=http", Some("192.0.2.43"))]
```
atimin, line 222: let's add a test for malformed first hop:

```
#[case("for=_hidden, for=198.51.100.17;proto=http", Some("198.51.100.17"))]
```

atimin (changes requested):
Hi @LordAizen1 ,

The PR looks great to me. Please update the changelog and the contribution guide accordingly. Also, add an additional test.

Thank you very much for your contribution!

reductstore/src/api/http/middleware.rs
LordAizen1: Thanks for the review, @atimin! I've added the malformed-first-hop test case and a CHANGELOG entry under Unreleased → Fixed. On "update the contribution guide", I wasn't 100% sure which you meant, so I added myself to the Contributors wall in the README. Is that what you had in mind, or were you referring to something specific in CONTRIBUTING.md? Happy to adjust to whichever you intended.

atimin: sorry, I wasn't clear.  I meant: "update the changelog accordingly the contribution guide".

Something like:

```
 Add a replication `compression` setting (`none`, `zstd`, `gzip`, default `none`) that compresses batch payloads during transfer using HTTP `Content-Encoding`, [Issue-1348](https://github.com/reductstore/reductstore/issues/1348) by @DibbayajyotiRoy
```

atimin (approved):
Perfect! Thanks you!

LordAizen1:
Thanks so much, @atimin! Appreciate the patient review and the quick merge. Happy I could contribute and hope to send a few more your way. 😄

atimin:
Your contribution is always welcome.  I label issues that can be completed by contributors.  You can also check out other projects like reduct-cli and web-console. If you have any questions, please ask the community at https://community.reduct.store/, 

Thanks again!

LordAizen1:
Thanks, appreciate it! I'll take a look at the labeled issues and at reduct-cli / web-console. Will likely be back with more soon. 😄

PR #1549, opened 2026-07-15
Add per-run limits to lifecycle policies

### LINKED ISSUE #1463: Feature: add per-run limits to lifecycle policies
## Summary
Lifecycle policies should allow configuring the per-run data processing interval independently from the task execution interval.

## Current behavior
Lifecycle progress is tracked through the system bucket with `last_processed_ts`. This approach is useful and should stay: it lets lifecycle policies process data again when data changes after a previous task run. For example, if a block was decompressed after the lifecycle task passed it, a later lifecycle run can compress it again.

The current implementation advances the processing window by a hard-coded multiplier of the lifecycle execution interval:

```rust
let data_window = interval_us.saturating_mul(24);
```

So a policy with `interval: "1h"` processes at most about `24h` of matching historical data per run.

## Problem
The processing interval is currently tied to the execution interval by a fixed `24x` multiplier.

That makes scheduling and catch-up behavior harder to tune independently. Operators may want the task to run often but process a smaller data interval per run, or run less often while still processing a larger data interval.

## Proposal
Add a lifecycle setting for the per-run data processing interval.

Proposed option name: `processing_interval`.

Example:

```json
{
  "type": "compress",
  "bucket": "data",
  "older_than": "30d",
  "interval": "1h",
  "processing_interval": "12h"
}
```

Semantics:

- `interval` controls how often the lifecycle task wakes up.
- `processing_interval` controls how much timestamp range the task may process in one run.
- If `processing_interval` is omitted, keep the current behavior for backward compatibility: `processing_interval = interval * 24`.

## Why this helps
- removes the hard-coded `24x` multiplier from lifecycle execution behavior
- lets operators tune task schedule and data processing window independently
- preserves the current system-bucket progress approach
- preserves current behavior by default

Harshal875:
Lifecycle policies previously used a fixed per-run data-time window of `24 * interval` when persisted lifecycle progress was available, and processed the full eligible range when system events were disabled. Large backlogs could therefore make a single run too expensive.

This PR adds one optional setting:

- `processing_interval`: maximum span of data time processed by one lifecycle run, for example `6h`, `12h`, or `1d`.

Behavior:

- If unset, existing behavior is preserved: lifecycle uses the default `24 * interval` data window.
- If set, `processing_interval` replaces the default data window when lifecycle progress is available.
- If system events are disabled, `processing_interval` still clamps the full-range processing window.
- The value is validated as a positive duration.
- The setting is available through the lifecycle API as `processing_interval`.
- The setting is available through provisioning as `RS_LIFECYCLE_<ID>_PROCESSING_INTERVAL`.
- The setting applies to both delete and compress lifecycle policies.

Follow-up SDK and public documentation task: #1577

Closes #1463

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Automatically create missing destination buckets during replication, [PR-1539](https://github.com/reductstore/reductstore/pull/1539) by @rohankumardubey
 - Add a replication `compression` setting (`none`, `zstd`, `gzip`, default `none`) that compresses batch payloads during transfer using HTTP `Content-Encoding`, [Issue-1348](https://github.com/reductstore/reductstore/issues/1348) by @DibbayajyotiRoy
+- Add a `processing_interval` setting to lifecycle policies that bounds how much data time a single run processes (defaults to `24 * interval` when unset), configurable via the API and `RS_LIFECYCLE_<ID>_PROCESSING_INTERVAL`, [PR-1549](https://github.com/reductstore/reductstore/pull/1549) by @Harshal875
 - Configure server-wide default bucket settings with `RS_DEFAULTS_BUCKET_*` environment variables, [PR-1535](https://github.com/reductstore/reductstore/pull/1535) by @rohankumardubey
 - Reject record writes before writing when the data folder filesystem has insufficient free disk space, in addition to the existing quota check, [PR-1525](https://github.com/reductstore/reductstore/pull/1525)
 - Support glob-like entry patterns in lifecycle task filters, including path-aware `*` and `**` wildcards and exclusion patterns, [PR-1526](https://github.com/reductstore/reductstore/pull/1526) by @upuddu
--- a/reductstore/src/lifecycle/action/progress.rs
+++ b/reductstore/src/lifecycle/action/progress.rs
@@ -27,8 +27,23 @@ pub(crate) async fn processing_window(
     let (first_record_start, effective_cutoff_stop) =
         matching_record_window(settings, context, cutoff_stop).await?;
 
+    let processing_interval_us = settings
+        .processing_interval
+        .as_ref()
+        .map(|interval| parse_duration_to_micros(interval))
+        .transpose()?
+        .map(|interval| interval.max(0) as u64);
+
     if !context.system_events_enabled {
-        let stop = effective_cutoff_stop;
+        // Without persisted progress, the processing interval still bounds
+        // each run: processed data becomes ineligible, so the window
+        // advances anyway.
+        let stop = match processing_interval_us {
+            Some(interval) => first_record_start
+                .saturating_add(interval)
+                .min(effective_cutoff_stop),
+            None => effective_cutoff_stop,
+        };
         return Ok(ProcessingWindow {
             start: Some(first_record_start.min(stop)),
             stop: Some(stop),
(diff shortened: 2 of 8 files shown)


reductstore/src/lifecycle/action/delete.rs
```diff
@@ -45,7 +47,7 @@ impl LifecycleAction for DeleteLifecycleAction {
             .get_bucket(&settings.bucket)
             .await?
             .upgrade()?;
-        let query_entry = QueryEntry {
+        let build_query = |entries: Option<Vec<String>>, limit: Option<u64>| QueryEntry {
```
atimin: Your implementation is correct, but it seems to only work for deletion because we can't use the 'when' condition for compression. I abandoned the idea a few days ago and updated the issue, but it looks like you implemented it for the old version.

reductstore/src/lifecycle/action/progress.rs
```diff
@@ -15,6 +15,10 @@ pub(crate) struct ProcessingWindow {
     pub(crate) start: Option<u64>,
     pub(crate) stop: Option<u64>,
     pub(crate) last_processed_ts: Option<u64>,
+    /// Progress persisted before this run. Reported instead of
+    /// `last_processed_ts` when a per-run record limit truncates the window,
+    /// so the next run re-covers the unfinished range.
+    pub(crate) prior_processed_ts: Option<u64>,
```
atimin: I don't understand why we need it. 
The initial idea was simply to make the data window adjustable.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Automatically create missing destination buckets during replication, [PR-1539](https://github.com/reductstore/reductstore/pull/1539) by @rohankumardubey
 - Add a replication `compression` setting (`none`, `zstd`, `gzip`, default `none`) that compresses batch payloads during transfer using HTTP `Content-Encoding`, [Issue-1348](https://github.com/reductstore/reductstore/issues/1348) by @DibbayajyotiRoy
+- Add per-run processing limits to lifecycle policies: `max_span_per_run` bounds how much data time one run processes and `max_records_per_run` caps removed records per run for delete policies, configurable via the API and `RS_LIFECYCLE_<ID>_MAX_SPAN_PER_RUN` / `RS_LIFECYCLE_<ID>_MAX_RECORDS_PER_RUN`, [ISSUE-1463](https://github.com/reductstore/reductstore/issues/1463)
```
atimin: It must be the PR ID not issue ID.  Please, add "by @Harshal875"

atimin (changes requested):
Hi @Harshal875 ,

Thank you for your contribution. I have a few remarks on it. If you are open to improvement, please read my comments.

I would appreciate it if you could create the PR in accordance with the CONTRIBUTING.md guide.

Thanks!

atimin:
Hi @Harshal875,
I see that you have marked my comments as 'Resolved', but I can't see any changes in the code. Did you forget to push something, or is it too early?

Harshal875:
Hi Yes, I am working on it.

Harshal875:
hi @atimin , 
Pushed the rework , replaced max_span_per_run/max_records_per_run with
a single processing_interval setting per the updated issue spec. Also
fixed the CHANGELOG entry (PR #1549 reference + credit). Ready for
another look whenever you have time.

atimin (approved):
Hi @Harshal875 ,

The PR is perfect now. I'm merging it. Thank you very much for your contribution!

Issue #1550, opened 2026-07-16
$system bucket should behave as a provisioned bucket


mother-6000:
## Summary
The `$system` bucket currently applies `RS_SYSTEM_EVENTS_QUOTA_SIZE` only when the bucket is created. After creation it behaves like a regular bucket, so users can remove it and startup does not reapply the configured quota.

## Expected behavior
`$system` should behave as a provisioned bucket managed by the server.

- The bucket is flagged/handled as provisioned.
- A user cannot permanently remove the system bucket.
- `RS_SYSTEM_EVENTS_QUOTA_SIZE` is applied on server startup/restart, not only during first creation.

## Actual behavior
`$system` receives `RS_SYSTEM_EVENTS_QUOTA_SIZE` only at creation time and can be removed by a user.

## Suggested fix
Mark `$system` as provisioned and apply the configured system events quota during startup/restart.

## Version
Affected: v1.20

Target fix: v1.21

LordAizen1:
Hi, I'd like to take this one as well if that's alright.

Planning to do it after #1551 so the default value lands first, since this one builds on top of it.

PR #1556, opened 2026-07-17
fix(cfg): default RS_SYSTEM_EVENTS_QUOTA_SIZE to 10 GB

### LINKED ISSUE #1551: RS_SYSTEM_EVENTS_QUOTA_SIZE should default to 10GB
## Summary
`RS_SYSTEM_EVENTS_QUOTA_SIZE` should have a default value of 10GB.

## Expected behavior
When `RS_SYSTEM_EVENTS_QUOTA_SIZE` is not explicitly configured, ReductStore uses a 10GB quota for system events.

## Actual behavior
The default value is missing or differs from the expected 10GB default.

## Version
Target fix: v1.21

LordAizen1:
Fixes #1551.

The `$system` bucket had no default quota, so leaving `RS_SYSTEM_EVENTS_QUOTA_SIZE` unset meant system events were stored without any quota. Every other setting in this module already has a `DEFAULT_SYSTEM_EVENTS_*` constant, this one was just missing.

Went with 10 GB in SI notation (10,000,000,000) as you said in the issue. That also matches how the rest of the code reads sizes: the bucket provisioning test asserts 1GB as `1_000_000_000`, and this module's own test asserts 10MB as `10_000_000`.

### Checked on a live server

Built and ran from this branch, no Docker image:

```
RS_DATA_PATH=./data cargo run -p reductstore --features "fs-backend"
```

With `RS_SYSTEM_EVENTS_QUOTA_SIZE` unset, `GET /api/v1/b/$system`:

```json
"settings": { "quota_type": "FIFO", "quota_size": 10000000000, ... }
```

Then restarted with `RS_SYSTEM_EVENTS_QUOTA_SIZE=5GB` to make sure an explicit value still wins:

```json
"settings": { "quota_type": "FIFO", "quota_size": 5000000000, ... }
```

### Tests

The two default tests were asserting `quota_size: None`, so they now assert the new default. `cargo test -p reductstore --lib cfg::` passes (125 tests).

Kept this to the default value only, #1550 stays separate as you suggested. I'll pick that one up next.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Default `RS_SYSTEM_EVENTS_QUOTA_SIZE` to 10 GB when it is not set, [Issue-1551](https://github.com/reductstore/reductstore/issues/1551) by @LordAizen1
 - Allow fork pull request CI to run without protobuf setup tokens or Docker Hub credentials by vendoring `protoc` and making Docker login optional, [PR-1532](https://github.com/reductstore/reductstore/pull/1532)
 - Pipelined replication batch sending to overlap preparing the next batch with sending the current one, [PR-1527](https://github.com/reductstore/reductstore/pull/1527) by @DibbayajyotiRoy (based on [PR-1415](https://github.com/reductstore/reductstore/pull/1415) by @FirasCh3)
 - Split block manager storage logic into smaller modules by responsibility, [PR-1503](https://github.com/reductstore/reductstore/pull/1503) by @rohankumardubey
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Default `RS_SYSTEM_EVENTS_QUOTA_SIZE` to 10 GB when it is not set, [Issue-1551](https://github.com/reductstore/reductstore/issues/1551) by @LordAizen1
```
atimin: it must be link to PR, see other records.

atimin (changes requested):
Hi @LordAizen1 ,

it's perfect, just a little issue in the changelog.

Thanks for your contribution!

atimin (approved):
Perfect!

PR #1557, opened 2026-07-18
fix(syslog): make $system a provisioned bucket and reapply its quota on restart

### LINKED ISSUE #1550: $system bucket should behave as a provisioned bucket

LordAizen1:
## Problem

Closes #1550.

Right now the `$system` bucket only picks up `RS_SYSTEM_EVENTS_QUOTA_SIZE` when
it's first created, and after that it acts like any normal bucket. That leaves
two problems:

- Anyone can delete `$system` through the API, so the server's own telemetry
  bucket isn't really protected.
- If you change `RS_SYSTEM_EVENTS_QUOTA_SIZE` and restart, nothing happens. The
  bucket just keeps whatever quota it had when it was first made.

## What this does

- `local_writer.rs`: when the writer creates `$system` on the first event, it
  now marks it provisioned, so it can't be removed, renamed or reconfigured
  through the API.
- `syslog.rs`: adds `provision_system_bucket`, which runs on startup. It re-arms
  the provisioned flag (that flag lives in memory, so it's gone after a reload)
  and re-applies the settings, so a changed quota actually takes effect.
  Replicas don't keep a local `$system`, so they get skipped.

## How I tested it

I ran a real server built from this branch with data on disk, then restarted it
with `RS_SYSTEM_EVENTS_QUOTA_SIZE` changed from 1GB to 2GB:

| | `is_provisioned` after restart | quota after restart | `DELETE /b/$system` |
|---|---|---|---|
| `main` | `false` | stuck at 1GB | 200, deleted |
| this branch | `true` | 2GB applied | 409, refused |

- Added a regression test `provisions_and_reapplies_quota_on_restart` that covers
  both parts: provisioned on creation, and the flag re-armed plus quota
  re-applied on restart.
- Ran `cargo fmt --all` and `cargo test -p reductstore --lib syslog::`, 71 passed.

## Notes

- No new config or feature flags. This only changes when the existing
  `RS_SYSTEM_EVENTS_QUOTA_SIZE` gets applied, now on every startup instead of
  just the first time.
- Updated `CHANGELOG.md` under Fixed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Treat the `$system` bucket as provisioned so it cannot be removed through the API, and reapply `RS_SYSTEM_EVENTS_QUOTA_SIZE` on startup instead of only at first creation, [Issue-1550](https://github.com/reductstore/reductstore/issues/1550) by @LordAizen1
 - Resolve client IP from multi-hop RFC 7239 `Forwarded` headers so token IP allowlists compare against the real client instead of the proxy, [PR-1546](https://github.com/reductstore/reductstore/pull/1546) by @LordAizen1
 - Normalize bucket history timestamps when only attachment metadata entries contain records, [PR-1534](https://github.com/reductstore/reductstore/pull/1534)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


atimin (approved):
It's perfect! Thank you!

PR #1567, opened 2026-07-19
Feat/replicate system events 1457

### LINKED ISSUE #1457: Bug: internal $system events are not replicated
## Summary
Internal `$system` events such as usage/statistics and audit records are not replicated even when a replication is configured with `$system` as the source bucket.

## Current behavior
A replication from `$system` to another bucket only replicates records that generate `TransactionNotification`s. User-facing write paths call `replication_repo.notify(...)`, but internal system loggers write directly through `storage.begin_write(...)` and bypass the replication notification path.

Examples affected:
- `$system/usage/...`
- `$system/audit/...`
- other internally emitted system/statistics events

## Expected behavior
When `$system` is configured as a replication source, internally written system events should produce replication transactions and be replicated like regular records.

## Root cause / implementation note
Replication is currently notification-driven, and notifications are emitted from API/write handlers instead of from a lower-level storage write path. We need to refactor replication notification so internal writes and API writes share the same notification mechanism, or otherwise add an explicit notification path for system logger writes.

## Target version
v1.21

DibbayajyotiRoy:
Closes #1457

Bug fix — internal `$system` events were silently not replicated even when `$system` was a configured replication source.

### What was changed?

Internal `$system` writes bypassed replication because only the API handlers call
`replication_repo.notify(...)`, while the system-event writer goes straight to
`StorageEngine::begin_write`. This PR connects the internal write path to replication:

- New `ReplicationNotifier` callback type in `syslog/sink.rs`, late-bound because the
  writer is constructed before the replication repo exists (same dependency shape as the
  log-sink registration from #1467).
- `LocalSystemLogger` builds a `TransactionNotification` (same struct and fields as the
  API handlers) after each successful write and invokes the registered notifier.
  Notification errors are logged and swallowed — a replication problem never fails a
  system-event write.
- The `replications/` and `logs/` families are **excluded** (matched on `SystemEventKind`,
  not entry strings) to prevent feedback loops: replication telemetry about the `$system`
  replication would feed it new transactions indefinitely, and captured replication
  error logs would do the same exactly during failures. Usage, audit, and lifecycle
  events all notify.
- Wiring in `cfg.rs` registers the notifier once the repo is built;
  `Components::replication_repo` becomes `Arc<AsyncRwLock<...>>` to share it with the
  callback (call sites unchanged via `Deref`).

API handler notify sites, the writer's persist/console paths, and replica-role
forwarding (which already notifies via the primary's HTTP API) are untouched.

Tests: `$system`-source replication gets exactly one pending transaction per usage
event; `replications/` and `logs/` events produce none; a failing notifier does not
fail the write; writer works with no notifier registered.

### Related issues

- Closes #1457
- Builds on the unified system-event writer from #1496 (single internal write site)
- Loop-guard rationale follows the reentrancy analysis from #1467

No. Users with a `$system`-source replication configured will start receiving usage,
audit, and lifecycle records on the destination (previously nothing was replicated);
`replications/` and `logs/` entries are intentionally not replicated.

### Other information:

Mechanism is "Plan B" approved in the issue discussion: notify from the single internal
writer rather than adding a storage-level hook, which would invert the storage→replication
layering and require refactoring all seven API notify sites.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Replicate internal `$system` events when `$system` is a replication source by notifying replication from the system-event writer; to prevent feedback loops, diagnostics of a `$system`-source replication and log messages from the replication module are excluded, [Issue-1457](https://github.com/reductstore/reductstore/issues/1457) by @DibbayajyotiRoy
+- Replicate internal `$system` events when `$system` is a replication source by notifying replication from the system-event writer; to prevent feedback loops, diagnostics of a `$system`-source replication and log messages from the replication module are excluded, [PR-1567](https://github.com/reductstore/reductstore/pull/1567) by @DibbayajyotiRoy
 - Treat the `$system` bucket as provisioned so it cannot be removed through the API, and reapply `RS_SYSTEM_EVENTS_QUOTA_SIZE` on startup instead of only at first creation, [PR-1557](https://github.com/reductstore/reductstore/pull/1557) by @LordAizen1
 - Resolve client IP from multi-hop RFC 7239 `Forwarded` headers so token IP allowlists compare against the real client instead of the proxy, [PR-1546](https://github.com/reductstore/reductstore/pull/1546) by @LordAizen1
 - Normalize bucket history timestamps when only attachment metadata entries contain records, [PR-1534](https://github.com/reductstore/reductstore/pull/1534)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Replicate internal `$system` events (usage, audit, lifecycle) when `$system` is a replication source by notifying replication from the system-event writer; the `replications/` and `logs/` families are excluded to prevent feedback loops, [Issue-1457](https://github.com/reductstore/reductstore/issues/1457) by @DibbayajyotiRoy
+- Replicate internal `$system` events when `$system` is a replication source by notifying replication from the system-event writer; to prevent feedback loops, diagnostics of a `$system`-source replication and log messages from the replication module are excluded, [Issue-1457](https://github.com/reductstore/reductstore/issues/1457) by @DibbayajyotiRoy
 - Treat the `$system` bucket as provisioned so it cannot be removed through the API, and reapply `RS_SYSTEM_EVENTS_QUOTA_SIZE` on startup instead of only at first creation, [PR-1557](https://github.com/reductstore/reductstore/pull/1557) by @LordAizen1
 - Resolve client IP from multi-hop RFC 7239 `Forwarded` headers so token IP allowlists compare against the real client instead of the proxy, [PR-1546](https://github.com/reductstore/reductstore/pull/1546) by @LordAizen1
 - Normalize bucket history timestamps when only attachment metadata entries contain records, [PR-1534](https://github.com/reductstore/reductstore/pull/1534)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Replicate internal `$system` events (usage, audit, lifecycle) when `$system` is a replication source by notifying replication from the system-event writer; the `replications/` and `logs/` families are excluded to prevent feedback loops, [Issue-1457](https://github.com/reductstore/reductstore/issues/1457) by @DibbayajyotiRoy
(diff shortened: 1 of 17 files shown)


DibbayajyotiRoy:
@atimin do have a review

reductstore/src/syslog/local_writer.rs
```diff
@@ -63,21 +71,242 @@ impl LocalSystemLogger {
...
+    /// and log records are node-local by design. A notification failure is
+    /// swallowed and logged — a replication problem must never fail a
+    /// system-event write.
+    async fn notify_replication(
+        &self,
+        event: &SystemEvent,
+        entry_name: String,
+        labels: reduct_base::Labels,
+    ) {
+        if matches!(
```
atimin: I read you comments on the infinite cycle problem.  You completely right, but I think we could work it around a bit better. 

We can add a flag to the system event:

```rust
 SystemEvent {                                                                                                                                                                                                    
         kind: SystemEventKind::Replication,                                                                                                                                                                          
         replicate,                                                                                                                                                                                                   
         // ...                                                                                                                                                                                                       
     }  
```

Then we could check the source bucket in ReplicationEventAggrigator and mark the replication as non-replicable. 

For logs, we could check the target from payload and filter out only errors coming from reductstore::replication module.

atimin (commented):
Hi @DibbayajyotiRoy , thank you very much for your PR.

I think we should work here a little bit more if you don't mind. Read my thoughts on the cycle replication and log problem. Morethrer, I found out an issue in the shutdown order and I have the following errors during the server shut down.


```
^C2026-07-19 12:28:37.762 (89120) [INFO] -- reductstore/launcher.rs:196 Received Ctrl-C, shutting down server...
2026-07-19 12:28:37.797 (91616) [WARN] -- reductstore/syslog/local_writer.rs:119 Failed to notify replication about system event 'total': [InternalServerError] Failed to enqueue replication notification
2026-07-19 12:28:37.797 (92864) [WARN] -- reductstore/syslog/local_writer.rs:119 Failed to notify replication about system event 'benchmark-1784303013882': [InternalServerError] Failed to enqueue replication notification
2026-07-19 12:28:37.798 (95360) [WARN] -- reductstore/syslog/local_writer.rs:119 Failed to notify replication about system event 'copy_system': [InternalServerError] Failed to enqueue replication notification
2026-07-19 12:28:37.798 (70528) [WARN] -- reductstore/syslog/local_writer.rs:119 Failed to notify replication about system event 'defaults': [InternalServerError] Failed to enqueue replication notification
2026-07-19 12:28:37.798 (81632) [WARN] -- reductstore/syslog/local_writer.rs:119 Failed to notify replication about system event 'dst': [InternalServerError] Failed to enqueue replication notification
2026-07-19 12:28:37.799 (99104) [WARN] -- reductstore/syslog/local_writer.rs:119 Failed to notify replication about system event 'src': [InternalServerError] Failed to enqueue replication notification
2026-07-19 12:28:37.799 (82880) [WARN] -- reductstore/syslog/local_writer.rs:119 Failed to notify replication about system event 'test': [InternalServerError] Failed to enqueue replication notification
2026-07-19 12:28:37.831 (56640) [INFO] -- reductstore/launcher.rs:191 Server has been shut down.
```

DibbayajyotiRoy:
Hi @atimin, thanks for the review. Both points are addressed.

Replicate flag: implemented as you suggested. SystemEvent now carries a serde-skipped replicate: bool (default true) and the writer checks it instead of matching on the event family:

ReplicationEventAggregator receives the replication's source bucket and marks its diagnostics non-replicable only when the source is $system. Diagnostics of ordinary replications now replicate, which the previous blanket family exclusion wrongly dropped.
LogCapture checks the snapshot's target and clears the flag only for records from reductstore::replication::*; all other captured logs replicate.
Shutdown warnings: the cause was ordering in StateKeeper::shutdown. Replication workers stop before the system-event collector, so the collector's final flushes were notifying a repo whose notification channel was already closed. Shutdown now detaches the notifier (set_replication_notifier(None)) before stopping the replication tasks, so those flushes skip notification cleanly instead of warning.

Added tests for the flag on both producers (source-bucket check, log-target filter), for the writer skipping non-replicable events, and for the detached notifier producing no transactions. Full suite is green.

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Replicate internal `$system` events when `$system` is a replication source by notifying replication from the system-event writer; to prevent feedback loops, diagnostics of a `$system`-source replication and log messages from the replication module are excluded, [Issue-1457](https://github.com/reductstore/reductstore/issues/1457) by @DibbayajyotiRoy
```
atimin: It must be a PR ID. Please make it shorter. After all, we can recover the context from the PR.

atimin (changes requested):
Hi @DibbayajyotiRoy , now it works perfectly!

Please update the changelog with the PR ID and remove the unnecessary comments. They are too long and often superficial. Only keep short descriptions for methods and new fields.

CHANGELOG.md
DibbayajyotiRoy: done!

atimin (approved):
Perfect, thank you!

Issue #363, opened 2023-10-07
Bug: 


atimin:
### What happened?

A bug happened!

### Version

1.4.0 (dev)

### OS

Linux

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #1572, opened 2026-07-22
Fix replication delete lock inversion

mother-6000:
Closes #

Bug fix.

### What was changed?

- Let replication removal and transaction notification use shared access to the outer replication repository trait object.
- Use read guards for API, Zenoh, and `$system` replication notification paths.
- Keep actual mutation protected by the internal replication-map locks.
- Add a regression test that logs a replicable `$system` event while an outer repository read guard is held, covering the lock pattern hit when deleting a replication task that flushes final diagnostics.

This prevents the DELETE /api/v1/replications/{name} path from holding an exclusive outer repository lock while a stopping replication task emits final `$system` diagnostics that notify replication.

### Related issues

Observed from website PR #363 CI: the documentation Go example `data_replication_remove.go` timed out deleting `repl-to-remove`, and server logs showed `$system` replication notification failing to acquire the async write lock.

No.

### Other information:

Validation:

- `cargo fmt --all`
- `cargo check --workspace`
- `cargo test -p reductstore syslog::local_writer::tests::usage_event_notifies_replication_while_repo_read_guard_is_held --features "fs-backend"`

Note: the first test rebuild hit `No space left on device`; I ran `cargo clean` in this worktree, which freed 5.6 GiB, then reran the validation successfully.

### DIFF
--- a/reductstore/src/syslog/local_writer.rs
+++ b/reductstore/src/syslog/local_writer.rs
@@ -132,15 +132,24 @@ impl LogSystemEvent for LocalSystemLogger {
 mod tests {
     use super::*;
     use crate::cfg::Cfg;
-    use crate::core::sync::AsyncRwLock;
+    use crate::core::sync::{reset_rwlock_config, set_rwlock_timeout, AsyncRwLock};
     use crate::replication::{ManageReplications, ReplicationRepoBuilder};
     use crate::syslog::{SystemEventKind, SYSTEM_BUCKET_NAME};
     use reduct_base::internal_server_error;
     use reduct_base::msg::replication_api::{ReplicationMode, ReplicationSettings};
     use rstest::{fixture, rstest};
     use serde_json::json;
+    use serial_test::serial;
     use std::sync::atomic::{AtomicUsize, Ordering};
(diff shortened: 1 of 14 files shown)


reductstore/src/syslog/local_writer.rs
```diff
@@ -132,16 +132,25 @@ impl LogSystemEvent for LocalSystemLogger {
...
     use crate::syslog::{SystemEventKind, SYSTEM_BUCKET_NAME};
     use reduct_base::internal_server_error;
     use reduct_base::msg::replication_api::{ReplicationMode, ReplicationSettings};
     use rstest::{fixture, rstest};
     use serde_json::json;
+    use serial_test::serial;
     use std::sync::atomic::{AtomicUsize, Ordering};
     use tokio::time::{sleep, Duration};
 
+    struct ResetRwLockConfig;
```
atimin: I'm not sure if this is correct.  We change the configuration to 1h at the shutdown stage. This is necessary for optimal performance in the cloud environment.  Here, we reverse this process.

atimin (changes requested):

atimin (approved):

PR #1573, opened 2026-07-23
require explicit API token auth configuration

### LINKED ISSUE #1501: Require explicit API token auth configuration for CRA compliance
## Summary

We need to tighten the token API / auth configuration flow in ReductStore to better support CRA compliance.

### 1) `RS_DISABLE_AUTH=true` explicitly disables authentication

Introduce / keep the rule:

- if `RS_DISABLE_AUTH=true`, authentication is disabled
- in that mode, token-related environment variables are ignored

This must be the **only** way to disable auth implicitly via configuration.

### 2) When auth is enabled, exactly one bootstrap token variable is required

When `RS_DISABLE_AUTH` is not set to `true`, authentication is enabled.
In that case:

- at least one bootstrap token variable is required
- the supported variables are:
  - `RS_API_TOKEN`
  - `RS_INIT_API_TOKEN`
- they **must not** both be set at the same time

So the valid authenticated startup modes are:

- `RS_API_TOKEN` set, `RS_INIT_API_TOKEN` unset
- `RS_INIT_API_TOKEN` set, `RS_API_TOKEN` unset

Invalid authenticated startup modes:

- neither variable set
- both variables set

### 3) `RS_API_TOKEN` provisions the first full-access token

`RS_API_TOKEN` must not be empty.

Behavior:

- provisions the first full-access token directly
- suitable for setups that want a fixed bootstrap/admin token from config

### 4) `RS_INIT_API_TOKEN` initializes the first full-access token

Behavior:

- initializes the first full-access token
- does **not** permanently provision or lock it
- after startup, the user can rotate it or remove it via the token API later, as long as the full-access-token safety rule is preserved

The goal is to support secure bootstrap without forcing a static token forever.

### 5) Prevent deletion of the last full-access token

Add validation in the token deletion API:

- deleting a token must be rejected if that would leave the system with **no remaining full-access token**

At least one token with full access must always exist while auth is enabled.

## Why

The current behavior is too implicit for compliance-sensitive deployments. We need:

- explicit opt-out for auth
- explicit bootstrap of an initial admin/full-access token
- mutually exclusive startup semantics
- protection against locking out administrative access
- clearer, auditable configuration behavior

## Acceptance criteria

- `RS_DISABLE_AUTH=true` disables auth and causes token env vars to be ignored
- when auth is enabled, exactly one of `RS_API_TOKEN` or `RS_INIT_API_TOKEN` must be set
- `RS_API_TOKEN` cannot be empty
- startup fails if both token vars are set
- startup fails if neither token var is set while auth is enabled
- `RS_INIT_API_TOKEN` creates an initial full-access token but does not prevent later rotation/removal
- token deletion API refuses deletion of the last remaining full-access token
- configuration and token API docs are updated accordingly
- tests cover startup validation, precedence rules, and token-deletion behavior

## Notes

Please confirm the final naming of `RS_INIT_API_TOKEN` (requested name) before implementation if there are any existing naming conventions that should override it.

rohankumardubey:
Closes #1501

Security improvement and breaking configuration change.

### What was changed?

- Added `RS_DISABLE_AUTH=true` as the explicit way to disable authentication.
- Require exactly one of `RS_API_TOKEN` or `RS_INIT_API_TOKEN` when authentication is enabled.
- Added `RS_INIT_API_TOKEN` for creating a rotatable and removable initial full-access token.
- Reject empty or conflicting bootstrap token configurations.
- Prevent deletion of the last full-access token.
- Recreate a removed `init-token` from `RS_INIT_API_TOKEN` on restart.
- Preserve the existing authentication behavior for read-only replicas.
- Mask bootstrap token values in configuration logs.
- Updated documentation, development commands, and CI test environments.
- Added tests covering configuration validation, token rotation and removal, restart behavior, and replicas.

### Related issues

https://github.com/reductstore/reductstore/issues/1501

Yes. Non-replica instances must now explicitly configure authentication by setting exactly one of `RS_API_TOKEN` or `RS_INIT_API_TOKEN`. To run without authentication, users must explicitly set `RS_DISABLE_AUTH=true`.

Read-only replica behavior remains unchanged.

### Other information:

Validation completed successfully:

- `cargo fmt --all -- --check`
- `cargo check --workspace`
- `cargo test --workspace --no-fail-fast -- --test-threads=1`

### DIFF
--- a/reductstore/src/cfg.rs
+++ b/reductstore/src/cfg.rs
@@ -88,6 +88,7 @@ pub struct Cfg {
     pub api_base_path: String,
     pub data_path: PathBuf,
     pub api_token: String,
+    pub api_token_is_provisioned: bool,
     pub cert_path: Option<PathBuf>,
     pub cert_key_path: Option<PathBuf>,
     pub ext_path: Option<PathBuf>,
@@ -124,6 +125,7 @@ impl Default for Cfg {
             api_base_path: "/".to_string(),
             data_path: PathBuf::from("/data"),
             api_token: "".to_string(),
+            api_token_is_provisioned: true,
             cert_path: None,
             cert_key_path: None,
             ext_path: None,
@@ -291,7 +293,8 @@ impl<EnvGetter: GetEnv, ExtCfg: ExtCfgBounds> CfgParser<EnvGetter, ExtCfg> {
         let lifecycles = Self::parse_lifecycles(&mut env);
         let has_lifecycles = !lifecycles.is_empty();
 
-        let api_token = env.get_masked("RS_API_TOKEN", "".to_string());
+        let (api_token, api_token_is_provisioned) =
+            Self::parse_auth_config(&mut env, ext_cfg.role());
 
         let cfg = Cfg {
             log_level: env.get("RS_LOG_LEVEL", DEFAULT_LOG_LEVEL.to_string()),
@@ -301,6 +304,7 @@ impl<EnvGetter: GetEnv, ExtCfg: ExtCfgBounds> CfgParser<EnvGetter, ExtCfg> {
             api_base_path,
             data_path: ext_cfg.data_path(),
             api_token: api_token.clone(),
+            api_token_is_provisioned,
             cert_path,
             cert_key_path,
             role: ext_cfg.role(),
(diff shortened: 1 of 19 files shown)


rohankumardubey:
@atimin  Its ready for review.

reductstore/src/cfg.rs
```diff
@@ -88,6 +88,7 @@ pub struct Cfg {
     pub api_base_path: String,
     pub data_path: PathBuf,
     pub api_token: String,
+    pub api_token_is_provisioned: bool,
```
atimin: i think we can do it more elegant with an enum:

```rust
enum ApiToken {
    NoToken,
    Provisioned(String),
    Initialized(String),
}
```

atimin (changes requested):
Hi @rohankumardubey , thank you for your contribution.  Your implementation is working perfectly, I  would only like to implement it with an enum instead of a boolean flag. I hope you don't mind.

Thank you!

reductstore/src/cfg.rs
rohankumardubey: That sounds like a better approach. I’ll update the implementation to use an enum. Thanks for the feedback!

atimin (approved):
Approved. Thanks for  your contribution!

PR #1591, opened 2026-08-06
feat(zenoh): key-prefix bucket routing and multiple key expressions

### LINKED ISSUE #1590: Zenoh API: route buckets by key prefix and accept multiple key expressions
**Problem**

The Zenoh API added in #1145 writes every sample into the single bucket named by `RS_ZENOH_BUCKET`, using the full key expression as the entry name.

In practice the Zenoh key space already carries a natural partition — per run, per cell, per site. Collapsing it into one bucket means quotas, lifecycle policies, and replication tasks can only be configured for the whole stream, not per partition. Retiring the data for one run, or replicating one cell's data to a different destination, means operating on the entry names inside a shared bucket rather than on a bucket boundary.

A second, related limitation: `RS_ZENOH_SUB_KEYEXPRS` and `RS_ZENOH_QUERY_KEYEXPRS` each accept exactly one key expression. Zenoh key expressions have no alternation operator, so two disjoint prefixes (say `sensors/**` and `events/**`) cannot be expressed as one expression. The only way to cover both today is to subscribe to `**` and take everything else with it.

**Proposed solution**

Add a bucket routing mode to the Zenoh API, selected by a new `RS_ZENOH_BUCKET_ROUTING` variable:

- `static` (default) — current behavior, everything to `RS_ZENOH_BUCKET`, full key as the entry name.
- `key-prefix` — the first chunk of the key expression selects the bucket, the remaining chunks become the entry name. `run_abc/motion/welder/commanded` becomes bucket `run_abc`, entry `motion/welder/commanded`.

The subscriber creates buckets on demand in `key-prefix` mode. Keys with no prefix to split on fall back to `RS_ZENOH_BUCKET`. The subscriber and the queryable resolve keys the same way, so what was written under a prefix is readable at the same key.

Separately, let `RS_ZENOH_SUB_KEYEXPRS` and `RS_ZENOH_QUERY_KEYEXPRS` accept a comma-separated list, declaring one subscriber/queryable per expression.

**Context**

Running a Zenoh-fronted ReductStore where the first key chunk is a run identifier. Per-run buckets make retention and replication configurable per run instead of per entry pattern.

**Alternatives**

- Run one ReductStore instance per partition — heavy, and the partition set is dynamic.
- Keep one bucket and express the partition through lifecycle/replication entry patterns — works for filtering, but quotas remain global to the bucket and the partition set has to be enumerated in policy config.
- A configurable chunk index or separator rather than "first chunk" — more flexible, but no use case yet that the first chunk does not cover.

**Acceptance criteria**

- `RS_ZENOH_BUCKET_ROUTING` supports `static` and `key-prefix`, defaulting to `static` so existing deployments are unaffected.
- `key-prefix` routes both the write path and the read path identically.
- Buckets are created on demand on the write path; the read path does not create them.
- `RS_ZENOH_SUB_KEYEXPRS` and `RS_ZENOH_QUERY_KEYEXPRS` accept a comma-separated list, with a single expression still valid.
- Unit tests for key resolution, config parsing, and subscriber routing behavior.
- `CHANGELOG.md` entry.

oh-yes-0-fps:
Resolves #1590.

## Problem

The Zenoh API writes everything into one bucket (`RS_ZENOH_BUCKET`), with the full key expression as the entry name. For deployments where the key space already carries a natural partition — per-run, per-cell, per-site — that forces all data into a single bucket, so quotas, lifecycle policies, and replication cannot be scoped per partition.

Separately, `RS_ZENOH_SUB_KEYEXPRS` / `RS_ZENOH_QUERY_KEYEXPRS` accept a single key expression. Zenoh key expressions have no alternation operator, so disjoint prefixes (`sensors/**` plus `events/**`) cannot be expressed without falling back to `**` and over-subscribing.

## Changes

**Key-prefix bucket routing.** New `RS_ZENOH_BUCKET_ROUTING`:

- `static` (default) — unchanged: everything goes to `RS_ZENOH_BUCKET`, full key as entry name.
- `key-prefix` — the first chunk of the key expression selects the bucket, the remainder becomes the entry name. `run_abc/motion/welder/commanded` → bucket `run_abc`, entry `motion/welder/commanded`.

Routing is a single `BucketRouting` type (`api/zenoh/routing.rs`) shared by the subscriber (write path) and queryable (read path), so both sides resolve keys identically. Leading and trailing slashes are trimmed. A single-chunk key (`orphan`) has no prefix to split on and falls back to `RS_ZENOH_BUCKET`.

In `key-prefix` mode the subscriber creates buckets on demand with default settings, memoizing the ones it has seen; a lost create race against another writer is tolerated. The queryable never creates buckets — a query against an unknown prefix returns the usual not-found error. Startup no longer pre-creates `RS_ZENOH_BUCKET` in this mode.

**Multiple key expressions.** `RS_ZENOH_SUB_KEYEXPRS` and `RS_ZENOH_QUERY_KEYEXPRS` now accept a comma-separated list, declaring one subscriber/queryable per entry. A single expression parses to a one-element list, so existing values behave as before.

## Configuration

| Variable | Values | Default |
| --- | --- | --- |
| `RS_ZENOH_BUCKET_ROUTING` | `static`, `key-prefix` | `static` |
| `RS_ZENOH_SUB_KEYEXPRS` | key expression, or comma-separated list | unset (write path disabled) |
| `RS_ZENOH_QUERY_KEYEXPRS` | key expression, or comma-separated list | unset (read path disabled) |

`RS_ZENOH_BUCKET_ROUTING` parsing is case-insensitive and accepts `key_prefix` / `keyprefix`; an unrecognized value falls back to `static`. `RS_ZENOH_BUCKET` keeps its meaning in `static` mode and becomes the fallback bucket in `key-prefix` mode.

## Compatibility

Defaults are unchanged, so existing deployments keep single-bucket behavior and their current key expressions. No API, storage, or wire-format changes.

## Validation

- `cargo fmt --all -- --check`, `cargo check --all-targets --all-features` — clean.
- `cargo test --all-features --lib` — 2295 passed. Run with `--test-threads=1`; the file-cache and block-manager tests contend on locks and time out under the default parallelism on this machine, independently of this change.
- New unit tests: key-to-`(bucket, entry)` resolution for both modes including slash trimming and single-chunk fallback, `RS_ZENOH_BUCKET_ROUTING` parsing and invalid-value fallback, and subscriber-level tests covering bucket creation on demand, fallback routing, and static mode leaving entry names untouched.
- Running `key-prefix` mode on a local build against a live Zenoh deployment for several weeks, with the first key chunk as a run identifier. Buckets are created as new prefixes appear, and the queryable reads back at the same keys the subscriber wrote.

## Changelog

Entry added under `### Added` in the unreleased section.

---

The text of this pull request description was drafted with an LLM. The code, the design decisions, and the validation above are mine, and I have reviewed everything I am submitting.


atimin:
Hi @oh-yes-0-fps, thank you very much for your contribution!

I think it definitely makes sense to add this feature to the ReductStore Zenoh API.

@AnthonyCvn, could you please take the lead on this?

Thank you!

AnthonyCvn:
Hi @oh-yes-0-fps thanks for working on this, I think it would be useful to understand the target use case a bit better before settling on the routing model.

Replication and lifecycle already support entry filtering, and lifecycle can also clean up data based on age, so I’m wondering which cases specifically require a separate bucket per key prefix. Is the main motivation per-bucket quotas/settings, or is there something else?

One thing I’m a little unsure about with `key-prefix` is that a single Zenoh subscriber/queryable can end up creating or accessing multiple ReductStore buckets dynamically based on key. That is different from the usual bucket-scoped access model that we have with the HTTP API.

Would an explicit configuration with multiple Zenoh subscribers/queryables, each mapped to a bucket, be an option? Something similar to `RS_ZENOH_SUB_<ID>_*` provisioning. If the buckets really need to be dynamic, it would be great to understand that requirement a bit more.

oh-yes-0-fps:
Hello Anthony,

After looking deeper it's likely the way I'm currently using ReductStore is not the ideally intended way.
I work at a startup that's doing high mix manufacturing using industrial robotic arms and there are discrete "runs" which are basically from the start of a specific assembly to the end. I originally had a custom solution with webtransport and custom http routing to write streamed binary blobs to a specific file that would then be served over http as well. I moved over to ReductStore and I kept this mentality of "1 run == 1 file ~= 1 bucket" and I currently have every run as it's own bucket.

You're right that entry filtering covers most of what I was after. Lifecycle `entries` takes globs and `older_than` handles age, so per-run cleanup never needed separate buckets. Worse, a lifecycle policy targets one literal bucket name, so with a bucket per run I can't express "delete runs older than 30 days" without creating a policy per run. I don't think my original usage should be encouraged.

I still think prefix routing is useful when the prefix set is stable and bounded:

- Data classes with different retention or quota in one key space, like high-rate raw vs low-rate derived. Different `quota_size` and different lifecycle policies, both per-bucket, and usually already split by the first key chunk.
- Per-site or per-cell aggregation where the first chunk is a site/cell id. Slowly changing, not always known when the config is written, and adding a cell shouldn't mean editing env vars and restarting.
- Buckets that differ by replication target, since `src_bucket` is one bucket per task.

On one subscriber touching many buckets, I think it's bigger than a style concern. The Zenoh runtime gets its components from `get_anonymous()` so it never goes through the token permission model, and `RS_ZENOH_BUCKET` is the only thing bounding which bucket it can reach today. In `key-prefix` mode the queryable resolves any key to a `(bucket, entry)` and calls `get_bucket` with no allowlist, so `RS_ZENOH_QUERY_KEYEXPRS=**` makes every bucket on the instance readable over Zenoh, including HTTP-created ones that have nothing to do with Zenoh.

Two things would fix it: gate auto-creation behind an explicit opt-in instead of implying it from the routing mode, and restrict routing to an allowlist or name pattern. One thing already works in your favor, created buckets pick up `cfg.bucket_defaults`, so `RS_DEFAULTS_BUCKET_*` governs them.

`RS_ZENOH_SUB_<ID>_*` is reasonable and matches `RS_BUCKET_<ID>_*` and `RS_LIFECYCLE_<ID>_*`, so it's the more consistent option. It costs config churn when a partition is added, which matters for the per-site case but not the data-class one. Happy to go that way, or make the explicit mapping the main path and keep `key-prefix` as a guarded opt-in. Tell me which you'd rather review and I'll rework it.

The multi-keyexpr change is independent of all this. Zenoh keyexprs have no alternation so two disjoint prefixes force `**` today.

AnthonyCvn:
@oh-yes-0-fps thanks for the detailed breakdown, this is helpful context.

I think the second option makes sense, so let's go with the explicit mapping as the main path and keep key-prefix as a guarded opt in.

Is the config below what you had in mind?

```
# fixed bucket, one block per bucket
RS_ZENOH_SUB_0_KEYEXPRS=raw/**
RS_ZENOH_SUB_0_BUCKET=raw_data

RS_ZENOH_SUB_1_KEYEXPRS=derived/**
RS_ZENOH_SUB_1_BUCKET=derived_data

# fixed bucket, several expressions into the same bucket
RS_ZENOH_SUB_2_KEYEXPRS=events/**,alarms/**
RS_ZENOH_SUB_2_BUCKET=events_data

# key-prefix, buckets come from the first chunk
RS_ZENOH_SUB_3_KEYEXPRS=site_$*/**,cell_$*/**
RS_ZENOH_SUB_3_ROUTING=key-prefix
RS_ZENOH_SUB_3_BUCKET_ALLOWLIST=site_*,cell_*
RS_ZENOH_SUB_3_ALLOW_BUCKET_CREATION=true

# read side mirrors the same shape
RS_ZENOH_QUERY_0_KEYEXPRS=raw/**
RS_ZENOH_QUERY_0_BUCKET=raw_data

RS_ZENOH_QUERY_1_KEYEXPRS=site_$*/**,cell_$*/**
RS_ZENOH_QUERY_1_ROUTING=key-prefix
RS_ZENOH_QUERY_1_BUCKET_ALLOWLIST=site_*,cell_*
```

Note that the allowlist filters bucket names with plain globs, while the field above takes Zenoh key expressions, so the syntax differs.

Would that solve what you are dealing with at the startup?

If you can work on this it would be great, I think it would be very beneficial for Zenoh users.

oh-yes-0-fps:
Hi, I am working on this, very busy time. I have done changes close to what you suggested in a staging branch in my fork but I want to get a chance to test it in my deployment before push it up for merging. Sorry for the delay!

I also have some performance improvements for value throughput I currently have deployed but I'm going to go over it again with my better understanding of ReductStore before I PR that.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Configure the Zenoh API as `RS_ZENOH_SUB_<ID>_*` and `RS_ZENOH_QUERY_<ID>_*` blocks, each declaring one or more key expressions (`_KEYEXPRS`, comma separated) and its own target bucket (`_BUCKET`); `_ROUTING=key-prefix` additionally lets the first chunk of a key expression select the bucket, bounded by `_BUCKET_ALLOWLIST` and, for subscribers, `_ALLOW_BUCKET_CREATION`. The unindexed `RS_ZENOH_BUCKET` / `RS_ZENOH_SUB_KEYEXPRS` / `RS_ZENOH_QUERY_KEYEXPRS` / `RS_ZENOH_QUERY_LOCALITY` settings keep working as a single static block when no indexed variable is set, [Issue-1590](https://github.com/reductstore/reductstore/issues/1590)
```
AnthonyCvn, line 17: You should set only one line in the change log with a link to this PR.

buildx.Dockerfile
```diff
@@ -12,7 +12,11 @@ RUN groupadd --gid 10001 reduct \
 
 RUN mkdir -p /data && chown 10001:10001 /data
 
-FROM debian:trixie-slim@sha256:28de0877c2189802884ccd20f15ee41c203573bd87bb6b883f5f46362d24c5c2
+FROM debian:trixie-slim@sha256:d7e12182ce18b85b93007c1dedf31f2d29e01ccf3182cc4017c709b6259bc132
+
+RUN apt-get update \
+    && apt-get upgrade -y --no-install-recommends \
+    && rm -rf /var/lib/apt/lists/*
```
AnthonyCvn, line 19: You probably added this because of the vulnerability error, right? You can ignore it and merge main into your branch. It's already solved there.

reductstore/src/cfg/zenoh.rs
```diff
@@ -119,43 +198,90 @@ impl FromStr for ZenohQueryableLocality {
...
+        let indexed_keys: Vec<&String> = zenoh_keys
+            .iter()
+            .filter(|key| is_indexed_block_key(key))
+            .collect();
+        let legacy_keys: Vec<&String> = zenoh_keys
+            .iter()
+            .filter(|key| LEGACY_KEYS.contains(&key.as_str()))
+            .collect();
+
+        let (subscribers, queryables) = if indexed_keys.is_empty() {
```
AnthonyCvn, line 263: A partial indexed configuration disables the existing unindexed subscriber but is then discarded. It would be better to fail startup instead of silently running Zenoh with no ingestion.

reductstore/src/api/zenoh/subscriber.rs
```diff
@@ -63,45 +87,133 @@ impl SubscriberPipeline {
...
+                )
+                .await
+            {
+                Ok(writer) => break writer,
+                // Publishers stamping faster than the microsecond resolution of a
+                // record timestamp land on an occupied slot; the next free one keeps
+                // the sample and its arrival order.
+                Err(err)
+                    if err.status() == ErrorCode::Conflict
+                        && ts.wrapping_sub(timestamp.unwrap_or(ts)) < MAX_TIMESTAMP_SHIFT_US =>
```
AnthonyCvn, line 120: This limit looks strange if zenoh has no timestamp it checks `ts - ts < MAX_TIMESTAMP_SHIFT_US`

reductstore/src/api/zenoh/session.rs
```diff
@@ -126,16 +155,127 @@ pub(crate) async fn run_session(
...
+                    kind, block.id, key_expr, err
+                ))
+            })?;
+        }
+    }
     Ok(())
 }
 
+// zenoh delivers a sample to every matching subscriber, so overlapping keyexprs store it twice
+fn warn_on_overlapping_keyexprs(config: &ZenohApiConfig) {
```
AnthonyCvn, line 212: I think this type of config keyexprs overlap should fail at startup:

 ```bash
RS_ZENOH_SUB_0_KEYEXPRS='sensors/**,sensors/camera/**'
RS_ZENOH_SUB_0_BUCKET='zenoh'
 ```

But this should remain valid:

 ```bash
RS_ZENOH_SUB_RAW_KEYEXPRS='sensors/**'
RS_ZENOH_SUB_RAW_BUCKET='raw'

RS_ZENOH_SUB_ARCHIVE_KEYEXPRS='sensors/camera/**'
RS_ZENOH_SUB_ARCHIVE_BUCKET='archive'
 ```

reductstore/src/api/zenoh/subscriber.rs
```diff
@@ -63,45 +87,133 @@ impl SubscriberPipeline {
...
+                {
+                    ts += 1;
```
AnthonyCvn, line 122: Is it ok to reject the write or do you need more than microsecond precision ?

AnthonyCvn (changes requested):
Hi @oh-yes-0-fps, thanks for the implementation. Could you also update the PR description to document the final block configuration?

PR #1594, opened 2026-08-07
Avoid blocking system-event replication on replication updates

### LINKED ISSUE #1593: Replication update can block system-event replication notification
## Description

Updating a replication task on `main` can cause `$system` event replication notification to time out with an async read-lock error.

Observed warning:

```text
2026-08-07 14:37:33.806 ( 3520) [WARN] -- reductstore/syslog/local_writer.rs:112 Failed to notify replication about system event 'test': [InternalServerError] Failed to acquire async read lock within timeout (caller: /home/runner/.cargo/git/checkouts/reductstore-0983dd190f4f924f/60ddf0e/reductstore/src/cfg.rs:460)
```

## What happens

The current `main` flow appears to be:

1. `PUT /api/v1/replications/:replication_name` acquires `components.replication_repo.write()` and holds it while awaiting `update_replication(...)`.
2. During the update, a `$system` event is written locally by `LocalSystemLogger`.
3. After the local write succeeds, `LocalSystemLogger` calls the replication notifier so the `$system` event is replicated.
4. The notifier registered in `cfg.rs` tries to acquire the same outer repository lock with `repo.read().await?.notify(notification).await`.
5. The read lock cannot be acquired while the update request still owns the write lock, so the notifier times out and logs the warning above.

## Impact

The `$system` event is persisted locally, but its replication notification is skipped. That means system events can be missed by replication while a replication task is being updated.

This seems specific to `main` after the unified `$system` logger / replication-notifier changes. Current `origin/stable` does not appear affected because it still uses `local_system_logger.rs`, which writes system events locally without the replication notifier callback.

## Expected behavior

Updating replication settings should not block or drop `$system` replication notifications. The replication update path should avoid holding the outer `components.replication_repo` write lock across awaited repository operations that can indirectly emit `$system` events.

## Possible fix

Reduce the scope/need for the outer repository write lock for normal replication CRUD operations. The repository already has internal locking around the replication task map, so methods such as `create_replication`, `update_replication`, `remove_replication`, and possibly `set_mode` could be adjusted to use `&self` where feasible. Then HTTP handlers can acquire `components.replication_repo.read()` for those operations, leaving the outer write lock for lifecycle-level operations that truly mutate the repository object, such as start/stop.

vjymisal0:
Closes #1593

Bug fix (lock ordering / concurrency).

### What was changed?

Confirmed the root cause in the actual code: `create_replication`, `update_replication`, `set_replication_provisioned`, and `set_mode` on `ManageReplications` all took `&mut self`, so the HTTP handlers had to call `components.replication_repo.write()` before invoking them. Their implementations, however, only mutate the repo's internal `Arc<AsyncRwLock<HashMap<String, ReplicationTask>>>` — the outer `&mut self` wasn't actually needed for these methods (only `start()`/`stop()` mutate plain fields and genuinely need it).

While the outer write lock was held across the awaited `update_replication(...)` call, a `$system` event written during that update tried to replicate itself through the notifier registered in `cfg.rs`, which does `repo.read().await?.notify(...)`. That read could never be granted while the write guard was alive, producing the async read-lock timeout from the issue.

This is the same pattern already applied to `remove_replication` in PR-1572 (shared/read access instead of exclusive access) — I extended it to the remaining CRUD methods:

- `reductstore/src/replication.rs`: changed `create_replication`, `update_replication`, `set_replication_provisioned`, `set_mode` to `&self` on the `ManageReplications` trait.
- `reductstore/src/replication/replication_repository/repo.rs`: updated the `ReplicationRepository` impl and its internal `create_or_update_replication_task` helper to `&self`.
- `reductstore/src/replication/replication_repository/read_only.rs`: updated the `ReadOnlyReplicationRepository` impl signatures to match.
- `reductstore/src/api/http/replication/{create,update,set_mode}.rs`: switched from `replication_repo.write()` to `replication_repo.read()`.
- `reductstore/src/cfg/provision/replication.rs`: dropped now-unnecessary `mut` bindings on the provisioning-time repo.
- `reductstore/src/syslog/local_writer.rs`: added `concurrent_replication_update_does_not_block_system_event_notification`, which drives `update_replication` through a read guard concurrently with a `$system` event write under a tight rwlock timeout (mirrors the existing `usage_event_notifies_replication_while_repo_read_guard_is_held` test). It also acts as a compile-time guard: if `update_replication` ever regresses to `&mut self`, calling it through a read guard stops compiling.
- `CHANGELOG.md`: added a `Fixed` entry under `Unreleased`.

`start()`/`stop()` still take `&mut self` and the HTTP layer / launcher still acquire `.write()` for those, since they mutate the repository's own `started`/`notification_worker` fields directly.

### Related issues

Closes #1593

No public API/behavior change. `ManageReplications` is `pub(crate)`-only surface (not exposed outside the crate), so the trait signature change is internal.

### Other information

I could not run `cargo fmt` / `cargo check` / `cargo test` locally — this environment has no Rust toolchain installed. I reviewed the diff manually, cross-checked every caller and both trait implementors (`ReplicationRepository`, `ReadOnlyReplicationRepository`) for the changed methods, and mirrored existing, already-compiling patterns in the same files (e.g. `remove_replication`/`notify` already using `&self` + `.read()`, and the existing rwlock-timeout test pattern in `syslog/local_writer.rs`) to minimize risk. Please let CI's `unit_tests` job be the actual pass/fail signal here — happy to fix anything it flags.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Avoid blocking `$system` event replication notifications while a replication task is being created, updated, or has its mode changed, by using shared outer replication repository access for those operations instead of exclusive access, the same pattern already used for removal, [Issue-1593](https://github.com/reductstore/reductstore/issues/1593)
 - Avoid a lock inversion when deleting a replication task that emits final `$system` diagnostics by using shared outer replication repository access for removal and transaction notifications, [PR-1572](https://github.com/reductstore/reductstore/pull/1572)
 - Replicate internal `$system` events when `$system` is a replication source by notifying replication from the system-event writer; to prevent feedback loops, diagnostics of a `$system`-source replication and log messages from the replication module are excluded, [PR-1567](https://github.com/reductstore/reductstore/pull/1567) by @DibbayajyotiRoy
 - Treat the `$system` bucket as provisioned so it cannot be removed through the API, and reapply `RS_SYSTEM_EVENTS_QUOTA_SIZE` on startup instead of only at first creation, [PR-1557](https://github.com/reductstore/reductstore/pull/1557) by @LordAizen1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Avoid blocking `$system` event replication notifications while a replication task is being created, updated, or has its mode changed, by using shared outer replication repository access for those operations instead of exclusive access, the same pattern already used for removal, [Issue-1593](https://github.com/reductstore/reductstore/issues/1593)
+- Avoid blocking `$system` event replication notifications while a replication task is being created, updated, or has its mode changed, by using shared outer replication repository access for those operations instead of exclusive access, the same pattern already used for removal, [PR-1594](https://github.com/reductstore/reductstore/pull/1594) by @vjymisal0
 - Avoid a lock inversion when deleting a replication task that emits final `$system` diagnostics by using shared outer replication repository access for removal and transaction notifications, [PR-1572](https://github.com/reductstore/reductstore/pull/1572)
 - Replicate internal `$system` events when `$system` is a replication source by notifying replication from the system-event writer; to prevent feedback loops, diagnostics of a `$system`-source replication and log messages from the replication module are excluded, [PR-1567](https://github.com/reductstore/reductstore/pull/1567) by @DibbayajyotiRoy
 - Treat the `$system` bucket as provisioned so it cannot be removed through the API, and reapply `RS_SYSTEM_EVENTS_QUOTA_SIZE` on startup instead of only at first creation, [PR-1557](https://github.com/reductstore/reductstore/pull/1557) by @LordAizen1
--- a/reductstore/src/syslog/local_writer.rs
+++ b/reductstore/src/syslog/local_writer.rs
@@ -287,21 +287,7 @@ mod tests {
         );
     }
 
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Avoid blocking `$system` event replication notifications while a replication task is being created, updated, or has its mode changed, by using shared outer replication repository access for those operations instead of exclusive access, the same pattern already used for removal, [Issue-1593](https://github.com/reductstore/reductstore/issues/1593)
```
atimin: Please,  link this PR not the issue. Add "by @vjymisal0" to be listed into the contributors.

reductstore/src/syslog/local_writer.rs
```diff
@@ -287,6 +287,73 @@ mod tests {
         );
     }
 
+    /// Regression test for #1593: updating a replication task must not hold
```
atimin: I would remove such a long comment. It's impractical to document every detail of the tests. The test has a good name, which is enough.

atimin (changes requested):
Hi @vjymisal0 ,  thank you very much for your contribution. 

The PR is almost perfect. Just needs some little changes. Read my comments, please.

CHANGELOG.md
vjymisal0: Fixed in c81bb52 — CHANGELOG now links PR-1594 and credits @vjymisal0.

reductstore/src/syslog/local_writer.rs
vjymisal0: Fixed in c81bb52 — trimmed the comment down to a one-line regression reference.

atimin (approved):
Perfect!  Thank you!

